Issue #8, opened 2022-06-02
[CI][API] Merge flake8 and api-test workflows into one


Samy-33:
Currently we have two workflows [api-tests](https://github.com/BitByte-TPC/gymkhana/blob/master/.github/workflows/api-tests.yml) and [api-flake8](https://github.com/BitByte-TPC/gymkhana/blob/master/.github/workflows/api_flake8.yml)

We can merge them together into a single workflow file with two jobs.

Read about Github actions [here](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions)

Start a discussion about your doubts [here](https://github.com/BitByte-TPC/gymkhana/discussions)

masterchief164:
I'd like to work on this issue

Samy-33:
Please go ahead, assigned it to you.

PR #20, opened 2022-06-04
refactor: merged api_tests and api_flake8 into api_build

masterchief164:
This PR is for the issue #8


.github/workflows/api-tests.yml
```diff
@@ -39,3 +39,19 @@ jobs:
...
+      uses: actions/checkout@v2
+    - name: Set up Python environment
+      uses: actions/setup-python@v2
+      with:
+        python-version: "3.8"
+    - name: flake8 Lint
+      uses: py-actions/flake8@v2
+      with:
+        max-line-length: "100"
+        path: "api"
```
Samy-33, line 57: I am not sure how `path` will work here when the `working directory` is already "api". Once your PR is approved and checks are run. Try to make sure it runs `flake8` on root `api` directory instead of `api/api`

Samy-33 (changes requested):
Rename the file: `api-build` or `api-ci` as it contains more than testing.

Also, refer to the [commit message convention](https://github.com/BitByte-TPC/gymkhana/blob/master/CONTRIBUTING.md#standard-commit-messages) that we need to follow, and change your message accordingly.

PR #39, opened 2022-06-05
ci(ui): fix failing ui action

Aksh-Bansal-dev:
The ui workflow was failing because node cache action couldn't find `package-lock.json`.

### DIFF
--- a/ui/src/dummy.test.ts
+++ b/ui/src/dummy.test.ts
@@ -3,3 +3,5 @@ describe('dummy test', () => {
     expect(1 + 2).toEqual(3);
   });
 });
+
+export {};
--- a/.github/workflows/ui-continuous_integeration.yml
+++ b/.github/workflows/ui-continuous_integeration.yml
@@ -5,10 +5,12 @@ on:
     branches: [master]
     paths:
       - ui/**
+      - .github/workflows/ui-continuous_integration.yml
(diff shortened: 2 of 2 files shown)


Samy-33:
Can you also please add the `ui-continuous-integration.yml` file to the paths to run this workflow for?

Aksh-Bansal-dev:
All checks are passing now. Should I change it back to `ui` directory?

Samy-33:
Workflow should run for the following cases:
- Any change in `ui` directory
- Any change in ui related workflows

ui/src/dummy.test.ts
```diff
@@ -3,3 +3,5 @@ describe('dummy test', () => {
     expect(1 + 2).toEqual(3);
   });
 });
+
+export {};
```
Samy-33, line 7: Is this required?

Aksh-Bansal-dev: Yes. Otherwise it gives this error: 
```
'dummy.test.ts' cannot be compiled under '--isolatedModules' because it is considered a global script file. Add an import, export, or an empty 'export {}' statement to make it a module.
```

This is just a dummy test and we can remove it once we start adding some real tests.

Samy-33 (approved):
LGTM

PR #50, opened 2022-06-06
ci(api): remove coverage commenter

Samy-33:
The coverage commenter action was introduced to comment out coverage of the API on the PR that was raised. This action requires the workflow to have write access to PRs.

The issue is in the forked repositories of users who don't have write access to `gymkhana` repository. When they raise a PR from their repo to `gymkhana` repo, workflows from their repositories are run and the `commenter action` in user repository workflow, does not have write access on original repo's PRs.

This causes workflow to fail with [error](https://github.com/BitByte-TPC/gymkhana/runs/6753223792?check_suite_focus=true)

### DIFF
--- a/.github/workflows/api-build.yml
+++ b/.github/workflows/api-build.yml
@@ -5,10 +5,12 @@ on:
     branches: [ master ]
     paths:
       - api/**
+      - .github/workflows/api-build.yml
   pull_request:
     branches: [ master ]
     paths:
       - api/**
+      - .github/workflows/api-build.yml
 
 env:
   PROJECT_SECRET_KEY: random-key
@@ -41,11 +43,6 @@ jobs:
     - name: Run Tests
       run: |
         pytest --cache-clear --cov=api | tee pytest-coverage.txt
-    - name: Pytest coverage comment
-      if: github.event_name == 'pull_request'
-      uses: MishaKav/pytest-coverage-comment@main
-      with:
-        pytest-coverage-path: api/pytest-coverage.txt
 
   flake8-lint:
     runs-on: ubuntu-latest


Aksh-Bansal-dev:
Maybe you can wrap the commit message line at 75 characters to pass the failing workflow.

Samy-33:
> Maybe you can wrap the commit message lines at 75 characters to pass the failing workflow.

Yup, I thought that I was writing a PR description instead of a commit :p

garg3133:
IIRC, we need to trigger the workflow on `pull_request_target` event instead of `pull_request` event. But there is some [security implication](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) in using `pull_request_target`.

The above article also describes a way we can combine `pull_request` event and `workflow_run` event to get the expected result.

Samy-33:
> IIRC, we need to trigger the workflow on `pull_request_target` event instead of `pull_request` event. But there is some [security implication](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) in using `pull_request_target`.

Thanks for sharing the above article.

Yes, using `pull_request_workflow` is a security concern that we shouldn't ignore just to have coverage report commented on the PR.

About using `pull_request` along with `workflow_run`, we need extra steps:
- In `pull_request` workflow, upload the `pytest-coverage.txt` file along with the `PR number`
- In `workflow_run` workflow, download the file (it gets downloaded as a zip, AFAIU from the article)
- Unzip the file
- Parse the `pytest-coverage.txt` ourselves into `markdown`
- Comment on the pull request using `PR number`

This sounds like a fairly complicated task to do. And I believe is not really worth it at the moment as we should focus on building the product. We can always manually check the coverage percentage from the `workflow` run and make sure, we ask the author of the PR to add more tests if coverage falls below say 95%.

Aksh-Bansal-dev (approved):
LGTM.

PR #76, opened 2022-06-17
feat(api): setup cors

varisha-025:


api/api/settings.py
```diff
@@ -43,18 +43,22 @@
...
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'corsheaders.middleware.CorsMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 ]
 
+CORS_ALLOWED_ORIGINS = os.environ.get('ALLOWED_HOSTS', 'http://127.0.0.1:8000').split(',')
```
Samy-33, line 60: Default should be empty string. AFAIK, we don't need the port in origins, do we?

Samy-33 (dismissed):
LGTM, just a minor change.

api/api/settings.py
varisha-025: Done.

PR #82, opened 2022-06-22
feat(api): add custom user model

divyanshkhetan:


Samy-33:
@divyanshkhetan @Saksham-1508, please fix the linting errors.

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
+import datetime
+from django.db import models
+from django.contrib.auth.models import AbstractUser, BaseUserManager, PermissionsMixin
+from django.contrib.auth import get_user_model
+
+# Create your models here.
```
Samy-33, line 6: this comment can be removed.

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+    TITLE = (
+        ('Prof', 'Prof'),
+        ('Dr', 'Dr'),
+        ('Mr', 'Mr'),
+        ('Mrs', 'Mrs'),
+        ('Ms', 'Ms'),
+    )
+
+
+class UserManager(BaseUserManager):
```
Samy-33, line 44: Is this a duplicate? We already have `CustomUserManager` in `managers.py`

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+        user.is_staff = True
+        user.is_superuser = True
+        user.save(using=self._db)
+        return user
+
+
+class User(AbstractUser, PermissionsMixin):
+    username = None
+    email = models.EmailField(unique=True)
+    gender = models.CharField(max_length=50, choices=Constants.SEX_CHOICES, default="M")
```
Samy-33, line 80: max_length 50 doesn't make sense. Keep it 10.

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+    is_active = models.BooleanField(default=True)
+    is_staff = models.BooleanField(default=False)
+    is_superuser = models.BooleanField(default=False)
+
+    USERNAME_FIELD = 'email'
+    REQUIRED_FIELDS = []
+
+    # user manager
+    objects = UserManager()
```
Samy-33, line 96: Should be `CustomUserModel`

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __str__(self):
+        return self.email
+
+    def save(self, *args, **kwargs):
+        """For cpanel."""
+        self.is_active = (self.is_active is True)
+        self.is_staff = (self.is_staff is True)
+        self.is_superuser = (self.is_superuser is True)
```
Samy-33, line 105: Q: What does `(self.is_active is True)` do?

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+    hostel_address = models.CharField(verbose_name='Hostel Address', max_length=200)
+
+    # bio
+    bio = models.TextField(max_length=1000, blank=True, null=True)
+
+    # socials
+    linkedin = models.URLField(null=True, blank=True, default="www.linkedin.com")
+    facebook = models.URLField(null=True, blank=True, default="www.facebook.com")
+    instagram = models.CharField(null=True, blank=True, max_length=1000)
+    github = models.URLField(null=True, blank=True, default="www.github.com")
```
Samy-33, line 125: There should be a separate model for these: `StudentSocial`. And default should be null.

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+    def __str__(self):
+        return self.user.email
+
+
+class Faculty(models.Model):
+    user = models.OneToOneField(get_user_model(), on_delete=models.CASCADE)
+    title = models.CharField(max_length=20, choices=Constants.TITLE)
+    department = models.CharField(max_length=100, choices=Constants.BRANCH)
+    designation = models.CharField(max_length=100, null=True)
+    # profile_img = models.ImageField(null=True)
```
Samy-33, line 136: Stale comment should be removed.

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __str__(self):
+        return self.user.email
+
+
+class Staff(models.Model):
+    user = models.OneToOneField(get_user_model(), on_delete=models.CASCADE)
+    # TODO: choices in department yet to be added
```
Samy-33, line 144: Create an issue and add the issue number in todo like: `TODO(#issueNumber): add choices in department`

api/api/accounts/models.py
```diff
@@ -0,0 +1,150 @@
...
+    department = models.CharField(max_length=100)
+    designation = models.CharField(max_length=100)
+    # profile_img = models.ImageField(null=True)
```
Samy-33, line 147: stale comment

Samy-33 (changes requested):

Samy-33:
Also, please remove empty files.

divyanshkhetan:
@Samy-33 sir, I have made the requested changes. Please review them. Thanks!

api/api/accounts/managers.py
```diff
@@ -0,0 +1,25 @@
+from django.contrib.auth.base_user import BaseUserManager
+
+
+class CustomUserManager(BaseUserManager):
+
+    def create_user(self, email, password, **extra_fields):
+        if not email:
+            raise ValueError('The Email must be set')
+        if not password:
+            raise ValueError("User must have a password")
```
Samy-33, line 10: Throwing this exception here means we are forcing users to have passwords. As we are using OAuth2 for now. That means we will not use passwords and so this check can be skipped.

api/api/accounts/models.py
```diff
@@ -0,0 +1,121 @@
...
+    department = models.CharField(max_length=100)
+    designation = models.CharField(max_length=100)
+
+    def __str__(self):
+        return self.user.email
+
+
+class StudentSocial(models.Model):
+    user = models.OneToOneField(get_user_model(), on_delete=models.CASCADE)
+    linkedin = models.URLField(null=True, blank=True, default=None)
```
Samy-33, line 114: I don't think we need to specific default in this case. If `null=True` then it'll automatically be populated as `None` by default.

Samy-33 (changes requested):

Samy-33 (approved):
LGTM

api/api/accounts/models.py
```diff
@@ -0,0 +1,119 @@
...
+    objects = CustomUserManager()
+
+    def __str__(self):
+        return self.email
+
+    def save(self, *args, **kwargs):
+        """For cpanel."""
+        self.is_active = (self.is_active is True)
+        self.is_staff = (self.is_staff is True)
+        self.is_superuser = (self.is_superuser is True)
```
Samy-33, line 72: I don't understand this and hence don't like it. Can you remove this?

Samy-33 (changes requested):

api/api/accounts/models.py
```diff
@@ -0,0 +1,115 @@
...
+    REQUIRED_FIELDS = []
+
+    # user manager
+    objects = CustomUserManager()
+
+    def __str__(self):
+        return self.email
+
+    def save(self, *args, **kwargs):
+        """For cpanel."""
```
Samy-33, line 69: Now we have this stale comment.

Samy-33 (changes requested):

Samy-33 (approved):

PR #84, opened 2022-06-23
feat(api): implement endpoint to create token

Samy-33:
This endpoint will be used when user signs up or logs in. Client
authenticates using Google and passes `id_token` to the API. API
verifies the `id_token`, extracts user information from it and generates
a token. If the user doesn't exist, it creates the user and sends back
user information with the token as well.

Current implementation has some issues:
 - If the user's token is expired, and user sends a request to create a
token, current implementation will return the expired token. It should
clean up the expired token, generate a new one and return it. An issue
should be created to fix this.
 - It has to be made sure that the client used to generate `id_token` was indeed the one configured for google oauth2.


Aksh-Bansal-dev (approved):

Aksh-Bansal-dev (approved):
LGTM.
Just squash all the commits.

api/api/auth/tests.py
```diff
@@ -0,0 +1,144 @@
...
+import pytest
+from django.contrib.auth import get_user_model
+from google.auth.exceptions import GoogleAuthError
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from api.auth.models import Token
+
+
+class TokenViewTest(APITestCase):
```
ankur-a, line 14: @Samy-33 tests can be moved to a separate tests folder.

Samy-33: Create an issue for this.

PR #86, opened 2022-06-24
feat(ui): add news carousel

### LINKED ISSUE #69: [UI] News carousel
The carousel on top.
 Design: [Figma](https://www.figma.com/file/GjOEDxaZ4sT9mBr8A10IiA/Gymkhana?node-id=2%3A2)

- Hide images on small screens, i.e. only display info card.
- All the data will come from a list. Each list element will have all the fields to display a carousel slide.

Aksh-Bansal-dev:
closes #69


Aksh-Bansal-dev:
@Samy-33 can you review this?

ui/src/components/NewsCarousel/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    clubName: 'The Programming Club',
+    description:
+      'Bayerische Motoren Werke AG, commonly referred to as BMW is a German multinational corporate manufactu',
+    imgUrl: dummyNewsImg,
+  },
+];
+
+export const NewsCarousel: React.FC = () => {
+  return (
+    <Carousel
```
Samy-33, line 32: When the data is empty it should should show no events yet at the center.

Samy-33 (changes requested):
Looks good overall, just a small change.

ui/src/components/NewsCarousel/index.tsx
Aksh-Bansal-dev: When the data is empty, this component does not render anything.

Samy-33: I see. Yeah, I think we can do away with this element if there are no highlighted events.

Samy-33 (approved):
LGTM.

PR #88, opened 2022-06-24
test(ui): add tests for Login component

### LINKED ISSUE #85: [UI] Add tests for Login
The tests should cover:
- Login component

Samy-33:
closes #85

### DIFF
--- a/ui/src/tests/pages/Login.test.tsx
+++ b/ui/src/tests/pages/Login.test.tsx
@@ -1,0 +1,109 @@
+import {createMemoryHistory} from 'history';
+import {createRoot, Root} from 'react-dom/client';
+import {act} from 'react-dom/test-utils';
+import {MemoryRouter, Router} from 'react-router-dom';
+
+import {Login} from '../../pages/Login';
+
+let container: Element | null = null;
+let root: Root | null = null;
+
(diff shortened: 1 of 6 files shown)


ui/src/tests/pages/Login.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    act(() => {
+      root!.render(
+        <MemoryRouter>
+          <Login />
+        </MemoryRouter>
+      );
+    });
+
+    const loginButton = document.querySelector('[data-testid=signin]');
+    expect(loginButton!.textContent!.trim()).toBe('SIGN IN');
```
Aksh-Bansal-dev, line 50: Do we need to check this here? This test should only check if the `sign in` button redirects to the google OAuth consent screen.

Aksh-Bansal-dev (changes requested):

ui/src/tests/pages/Login.test.tsx
Samy-33: You're right. Added a separate test for checking if the component renders properly.

Aksh-Bansal-dev (approved):
LGTM.

PR #94, opened 2022-06-27
fix(global): use authorization_code instead of id_token for user login

Samy-33:


api/api/auth/views.py
```diff
@@ -20,9 +22,19 @@ def post(self, request):
...
+                auth_response = google_client.exchange_auth_code_for_token(code)
+            except Exception:
+                logging.error('Error while exchanging auth code for token')
+                return Response(
+                    status=403,
+                    data={'message': 'Invalid authorization code. Try logging in again.'})
+            try:
+                # If the verification is done too soon, an error is thrown.
+                # Following 2 seconds pause is used to avoid that error.
+                time.sleep(2)
```
Aksh-Bansal-dev, line 36: why do we need to sleep for 2 seconds

ui/src/api/googleOauth.ts
```diff
@@ -28,31 +22,28 @@ export const openGoogleOAuthPage = () => {
...
+ * @param authorizationCode - Authorization code returned by google
  * @param redirectToHome - callback function to redirect to home
  */
 export const convertAccessToken = async (
-  googleAccessToken: string,
+  authorizationCode: string,
   redirectToHome: () => void,
   failRedirect: () => void
 ) => {
+  console.log(authorizationCode);
```
Aksh-Bansal-dev, line 33: Remove console.log

ui/.env.template
```diff
@@ -1,5 +1,3 @@
 VITE_GOOGLE_CLIENT_ID=
-VITE_API_CLIENT_ID=
-VITE_API_CLIENT_SECRET=
```
Aksh-Bansal-dev, line 3: Also, remove these from `constants.ts`

Aksh-Bansal-dev (changes requested):
Maybe `global` scope would be better in the commit message.

api/api/auth/views.py
Samy-33: Explained in the comments

ui/.env.template
Samy-33: Already removed.

ui/src/api/googleOauth.ts
Samy-33: This should probably be caught in linting test. Can you figure out if there's a way?

Aksh-Bansal-dev: Yeah, we can add this in `.eslintrc.js`.
```javascript
...
  rules: {
    'no-console': 2,
...
```

Samy-33: Create an issue for this.

Aksh-Bansal-dev (approved):
LGTM.

PR #111, opened 2022-07-04
feat(api): add model for storing clubs information

PranshuNayak:


api/api/clubs/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class ClubsConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
+    name = 'api_clubs'
```
Samy-33, line 6: name should be `api.clubs`
Look at the old apps for similarity.
For example: https://github.com/BitByte-TPC/gymkhana/blob/master/api/api/accounts/apps.py

api/api/clubs/models.py
```diff
@@ -0,0 +1,19 @@
...
+
+class Club (models.Model):
+    CULTURAL_CLUB = 'Cultural'
+    SCIENCE_AND_TECH_CLUB = 'Science & Technology'
+    SPORTS_CLUB = 'Sports'
+
+    CLUB_TYPE_CHOICES = [
+        ('Cultural', CULTURAL_CLUB),
+        ('S&T', SCIENCE_AND_TECH_CLUB),
+        ('Sports', SPORTS_CLUB)
```
Samy-33, line 12: Is there an upside to keeping these as variables? `CULTURAL_CLUB`, `SCIENCE_AND...` etc.

Samy-33 (changes requested):

api/api/clubs/models.py
PranshuNayak: I suppose use of variables increases code readability and provides single point of change if necessary.

Samy-33: For readability: CULTURAL_CLUB as a string is also pretty readable

But yes, single point of change in variables is useful.

PranshuNayak: So should I change or keep the variables .

Samy-33 (approved):

PR #118, opened 2022-07-07
feat(api): add model for club roles

PranshuNayak:


PranshuNayak:
@Samy-33 raised the PR

api/api/Roles/models.py
```diff
@@ -0,0 +1,22 @@
...
+
+    CLUB_ROLE_TYPES = [
+        ('Coordinator', 'Coordinator'),
+        ('Co-Coordinator', 'Co-Coordinator'),
+        ('Faculty Incharge', 'Faculty Incharge'),
+        ('Convener', 'Convener'),
+        ('Co-convener', 'Co-convener'),
+        ('Conselor', 'Conselor'),
+        ('Associate-Conselor', 'Associate-Conselor')
+    ]
```
Aksh-Bansal-dev, line 16: We should also add a role for core members as it is not necessary that only the coco or the coordinator will host an event.

api/api/Roles/models.py
```diff
@@ -0,0 +1,23 @@
...
+
+    name = models.CharField(max_length=30, choices=CLUB_ROLE_TYPES)
+    club = models.OneToOneField(Club, on_delete=models.CASCADE)
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
```
Samy-33, line 21: Why are these `OneToOneFields`? A club can have multiple roles, and a user can have multiple roles as well.

Samy-33 (changes requested):

api/api/Roles/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class EventsConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
+    name = 'api.Roles'
+    label = 'api_Roles'
```
Samy-33, line 7: name and label must be lowercase.

Samy-33 (changes requested):

api/api/Roles/models.py
```diff
@@ -0,0 +1,23 @@
...
+        ('Faculty Incharge', 'Faculty Incharge'),
+        ('Convener', 'Convener'),
+        ('Co-convener', 'Co-convener'),
+        ('Conselor', 'Conselor'),
+        ('Associate-Conselor', 'Associate-Conselor')
+    ]
+
+    name = models.CharField(max_length=30, choices=CLUB_ROLE_TYPES)
+    club = models.ManyToManyField(Club, on_delete=models.CASCADE)
+    user = models.ManyToManyField(User, on_delete=models.CASCADE)
```
Samy-33, line 21: These should be `ForeignKeyFields`. Why?

club as a `ManyToManyField` means, 1 object of club will have multiple objects of `Role` class, and 1 object of `Role` will have multiple clubs associated with it.

Do we want that? No.

One object of `Role` class means: one relationship between club, user and the role_name. Hence `ForeignKeyField` makes more sense.

Samy-33 (changes requested):

Samy-33 (approved):
LGTM

PR #120, opened 2022-07-07
feat(api): add event model

PranshuNayak:
@Samy-33 raised PR

### DIFF
--- a/api/api/events/models.py
+++ b/api/api/events/models.py
@@ -1,0 +1,18 @@
+from accounts.models import User
+from clubs.models import Club
+from django.db import models
+
+
+class Events(models.Model):
+    name = models.CharField(max_length=100)
+    description = models.TextField()
+    starts_at = models.DateField()
+    ends_at = models.DateField()
+    registration_required = models.BooleanField(default=False)
+    created_by = models.ForeignKey(
+        User, null=True, related_name='events_created', on_delete=models.SET_NULL)
+    image_url = models.URLField(max_length=500)
+    # This club is the primary organizer of this event
+    club = models.ForeignKey(Club, related_name='events', on_delete=models.PROTECT)
+    registration_approval_required = models.BooleanField(default=False)
+    address = models.CharField(max_length=500)
--- a/api/api/events/apps.py
+++ b/api/api/events/apps.py
@@ -1,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class EventsConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
+    name = 'api.events'
+    label = 'api_events'


api/api/events/models.py
```diff
@@ -0,0 +1,16 @@
...
+from django.db import models
+
+
+class Events(models.Model):
+    name = models.CharField(max_length=100)
+    description = models.TextField()
+    starts_at = models.DateField()
+    ends_at = models.DateField()
+    registration_required = models.BooleanField(default=False)
+    created_by = models.OneToOneField(User, on_delete=models.PROTECT)
```
Samy-33, line 12: `created_by` should be set `null` if the user is deleted. For that you'd need to make this field nullable as well.

api/api/events/models.py
```diff
@@ -0,0 +1,16 @@
...
+    image_url = models.URLField(max_length=500)
+    club = models.OneToOneField(Club, on_delete=models.PROTECT)
```
Samy-33, line 14: Add a comment on top of this: `This club is the primary organizer of this event.`

api/api/events/models.py
```diff
@@ -0,0 +1,16 @@
...
+    registration_approval_required = models.BooleanField(default=False)
+    address = models.TextField()
```
Samy-33, line 16: Why can't it be a `CharField`? `max_length` could be 500.

Samy-33 (changes requested):

Samy-33:
When you need the PR to be reviewed, don't leave it in `Draft` mode. Mark it ready for review.

api/api/events/models.py
```diff
@@ -0,0 +1,17 @@
...
+from django.db import models
+
+
+class Events(models.Model):
+    name = models.CharField(max_length=100)
+    description = models.TextField()
+    starts_at = models.DateField()
+    ends_at = models.DateField()
+    registration_required = models.BooleanField(default=False)
+    created_by = models.OneToOneField(User, null=True, on_delete=models.SET_NULL)
```
Samy-33, line 12: `created_by` should be a `ForeignKeyField`.

Add `related_name` to all the `ForeignKeys`, `OneToOne` and `OneToMany` relationships.

api/api/events/models.py
```diff
@@ -0,0 +1,17 @@
...
+    image_url = models.URLField(max_length=500)
+    # This club is the primary organizer of this event
+    club = models.OneToOneField(Club, on_delete=models.PROTECT)
```
Samy-33, line 15: should be a `ForeignKeyField`

Samy-33 (changes requested):

api/api/events/models.py
PranshuNayak: for `created_by` related name `events_created` works ?

Samy-33 (approved):
LGTM

PR #112, opened 2022-07-04
Add fixture for superuser

brijsiyag:
Fixes part of : #75 

Added fixture for superuser with following data - 
1. email - admin@iiitdmj.ac.in, password - Admin@gymkhana
2. email - adminstaff@iiitdmj.ac.in, password - Admin@gymkhana

PR #122, opened 2022-07-08
feat(api): add endpoint to get all clubs details

PranshuNayak:


Aksh-Bansal-dev (commented):
Can you also add some initial data? Maybe add a few clubs from each category.

api/api/clubs/migrations/0001_initial.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 4.0.5 on 2022-07-08 10:39
```
Samy-33, line 1: remove all the migration files. Once we are confident about all the db schemas, we'll push migrations together in one go.

api/api/clubs/views.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from .models import Club
+from .serializers import ClubsSerializer
+
+
+@api_view()
+def clubs_details(request):
```
Samy-33, line 9: Let's prefer class based views for consistency.
https://github.com/BitByte-TPC/gymkhana/blob/accc96ee7be245e82536f3bc7657e9ceb183c7c6/api/api/auth/views.py#L18
https://www.django-rest-framework.org/api-guide/views/#class-based-views

Samy-33:
> Can you also add some initial data? Maybe add a few clubs from each category.

For reference: #112

api/api/clubs/views.py
```diff
@@ -0,0 +1,12 @@
...
+
+from .models import Club
+from .serializers import ClubsSerializer
+
+
+class ClubsDetails(APIView):
+    def get(self, request):
+        queryset = Club.objects.all()
+        serializer = ClubsSerializer(queryset, many=True)
+        return Response(serializer.data)
```
Samy-33, line 12: I believe, we won't need this method if we extend the `ListApiView`

This should do:
```python
class ListClubsView(ListApiView):
    queryset = Clubs.objects.all()
    serializer_class = ClubsSerializer
```

Samy-33 (changes requested):
Also please add tests for the list view.

They should cover some edgecases like:
- When there's no clubs in db, return empty list.
- When there's n clubs in db, return n clubs in a list.

PranshuNayak:
> Also please add tests for the list view.
> 
> They should cover some edgecases like:
> 
>     * When there's no clubs in db, return empty list.
> 
>     * When there's n clubs in db, return n clubs in a list.

Not 

> Also please add tests for the list view.
> 
> They should cover some edgecases like:
> 
>     * When there's no clubs in db, return empty list.
> 
>     * When there's n clubs in db, return n clubs in a list.

from django.urls import reverse
from rest_framework.test import APITestCase
from api.clubs.models import Club

class ListClubViewsTest(APITestCase):
    def test_club_views_GET(self):
        url = '/clubs/'
        response = self.client.get(url,format='json')
        self.assertEqual(Club.objects.count(),len(response.data))
here club.objects.count() returns 0
What am I doin wrong ?

api/api/clubs/tests.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework.test import APITestCase
+
+from api.clubs.models import Club
+
+
+class ListClubViewTest(APITestCase):
+    """Tests ListClubView for getting the list of clubs."""
+
+    def test_listclubview_no_club_exist_return_empty_list(self):
```
Samy-33, line 9: Naming of the testcase methods:
1. Start with `test{featureBeingTested}
2. Multiple words in a sentence in a camelCase.
3. Separate sentences by an underscore.

So this becomes:
`testListClubs_noClubExistsInDb_returnsEmptyList`

api/api/clubs/tests.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework.test import APITestCase
+
+from api.clubs.models import Club
+
+
+class ListClubViewTest(APITestCase):
```
Samy-33, line 6: `ListClubsViewTest`

api/api/clubs/tests.py
```diff
@@ -0,0 +1,22 @@
...
+    """Tests ListClubView for getting the list of clubs."""
+
+    def test_listclubview_no_club_exist_return_empty_list(self):
+        url = '/clubs/'
+        response = self.client.get(url, format='json')
+        self.assertEqual(Club.objects.count(), len(response.data))
+
+    def test_listclubview_countOfClubsinDatabase_equals_countOfClubs_returned(self):
```
Samy-33, line 14: modify similar to above suggestion.

Samy-33 (changes requested):

api/api/clubs/tests.py
PranshuNayak: This works `testListClubs_countOfClubsInDb_returnsEqualCountOfClubs`  ?

Samy-33: `testListClubs_oneClubExistsInDb_returnsOneClubInResponse`

Make what the testcase checks for painfully clear to the reader.

api/api/urls.py
```diff
@@ -10,6 +10,7 @@
     path('api-auth/', include('rest_framework.urls', namespace='rest_framework')),
     path('auth/', include('api.auth.urls')),
     path('ping/', ping_handler, name='ping'),
+    path('clubs/', include('api.clubs.urls'))
```
Samy-33, line 13: Let's keep these paths in alphabetical order. It would be easy for others if they're trying to find a path.
```suggestion
    path('clubs/', include('api.clubs.urls')),
    path('ping/', ping_handler, name='ping'),
```

api/api/clubs/tests.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework.test import APITestCase
+
+from api.clubs.models import Club
+
+
+class ListClubsViewTest(APITestCase):
+    """Tests ListClubView for getting the list of clubs."""
+
+    def testListClubs_noClubExistsInDb_returnsEmptyList(self):
+        url = '/clubs/'
```
Samy-33, line 10: 1. `url` can be inlined.
2. add `given, when, then`.

For reference look at this file:
https://github.com/BitByte-TPC/gymkhana/blob/a1c20855145975831bfc892ffcafe33f2d137ae4/api/api/auth/tests.py#L27

api/api/clubs/tests.py
```diff
@@ -0,0 +1,22 @@
...
+        response = self.client.get(url, format='json')
+        self.assertEqual(Club.objects.count(), len(response.data))
+
+    def testListClubs_oneClubExistsInDb_returnsOneClubInResponse(self):
+        url = '/clubs/'
```
Samy-33, line 15: This can be inlined.

`response = self.client.get('/clubs/'...)`

Samy-33 (changes requested):

PranshuNayak:
The `API Tests / build (3.8) (pull_request)` is failing due to permission errors. `/clubs/` endpoint returns `Forbidden`

Samy-33:
> The `API Tests / build (3.8) (pull_request)` is failing due to permission errors. `/clubs/` endpoint returns `Forbidden`

It should ideally throw `UNAUTHORIZED` instead of `Forbidden`. Let me check, there's an issue with authentication middleware.

PranshuNayak:
@Samy-33 unable to understand how to write test for this protected route.

Samy-33:
> unable to understand how to write test for this protected route.

You can force authenticate a client:
```python
self.client.force_authenticate(user)
```

Then access protected endpoints using this client.

PranshuNayak:
@Samy-33 made the changes

PranshuNayak:
A problem 'messed up git history ' and rebasing is not supporting for `commit 92ee687e4d26843dd1df49125341be97a47b1101 ` I.e. the 2nd commit for this branch . So is there a way out or I should close this and create a new pull request

PR #127, opened 2022-07-08
feat(api): add bash script to make migrations

### LINKED ISSUE #123: [API] Add an sh script to generate migrations in correct order
In multi-app and multi-model django application, migrations are a way to keep track of db schema changes.

And our apps are dependent on each other and ideally should form a Directed Acyclic Graph of dependencies between them.

Everytime the develop needs to run makemigrations (be it first time setup or docker container setup), they should not have to worry about the order in which they have to `makemigrate` the apps.

A bash script that has this order defined will help the developers to skip this highly confusing step to build the app.

- Create a directory called `scripts` in the API `root`.
- Call this script `makemigrations.sh`
- It should be executable
- It should run makemigrations for all the apps that have models, in correct order.

arminpatel:
fixes #123

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -24,7 +24,7 @@ RUN pip install --upgrade pip && pip install -r requirements.txt
 COPY . $API_HOME  
 
 # make migrations
-RUN python manage.py makemigrations && python manage.py migrate
+RUN chmod +x scripts/makemigrations.sh && ./scripts/makemigrations.sh && python manage.py migrate
 
 # port where the Django app runs  
 EXPOSE 8000  
--- a/api/README.md
+++ b/api/README.md
@@ -12,7 +12,7 @@ Make sure you are using `python 3.8` as we are officially working on it. If some
 - Create a virtual environment: `$ python -m virtualenv .venv`
 - Activate the virtual environment: `$ source .venv/bin/activate` (On windows: `> ./.venv/Scripts/activate`)
 - Install the dependencies: `$ pip install -r requirements.txt`
-- Make Migrations: `$ python manage.py makemigrations` (This commands generates migrations which would be used to update database schemas)
+- Make Migrations: `$ chmod +x scripts/makemigrations.sh && sh ./scripts/makemigrations.sh` (This commands generates migrations which would be used to update database schemas)
 - Migrate the DB: `$ python manage.py migrate`
 - Run server: `$ python manage.py runserver`
 - Run tests: `$ pytest`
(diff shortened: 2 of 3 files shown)


Aksh-Bansal-dev (commented):
Make necessary changes in the Dockerfile.

arminpatel:
> Make necessary changes in the Dockerfile.

done

api/scripts/makemigrations.sh
```diff
@@ -0,0 +1,5 @@
+#!bin/bash
```
Samy-33, line 1: It should probably be `#!/bin/bash`

Try executing it: `./scripts/makemigrations.sh`

api/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN pip install --upgrade pip && pip install -r requirements.txt
 COPY . $API_HOME  
 
 # make migrations
-RUN python manage.py makemigrations && python manage.py migrate
+RUN sh scripts/makemigrations.sh && python manage.py migrate
```
Samy-33, line 27: `makemigrations.sh` should be executed independently by `./scripts/makemigrations`

That's how we are expecting it to be run in local and deployed env. The behaviour should be consistent.

Samy-33 (changes requested):

api/scripts/makemigrations.sh
arminpatel: > It should probably be `#!/bin/bash`
yeah 😅

api/README.md
```diff
@@ -12,7 +12,7 @@ Make sure you are using `python 3.8` as we are officially working on it. If some
 - Create a virtual environment: `$ python -m virtualenv .venv`
 - Activate the virtual environment: `$ source .venv/bin/activate` (On windows: `> ./.venv/Scripts/activate`)
 - Install the dependencies: `$ pip install -r requirements.txt`
-- Make Migrations: `$ python manage.py makemigrations` (This commands generates migrations which would be used to update database schemas)
+- Make Migrations: `$ sh scripts/makemigrations.sh` (This commands generates migrations which would be used to update database schemas)
```
Samy-33, line 15: Update the doc.

Samy-33 (changes requested):
Make sure the script file is executable `chmod`
And the Dockerfile runs as expected.

arminpatel:
Changed the script to execute independently and accordingly changed the doc. Docker image builds correctly.

Samy-33 (approved):
LGTM

PR #131, opened 2022-07-09
feat(api): add auth middleware

### LINKED ISSUE #97: Middleware to add user info in request if the token is available in headers
After a request is received by our django server, and before it reaches the `view` code to be executed, the user information needs to be added in the request.

We do that by fetching user for the token available in the request headers and populating the request in a middleware.

Read more about [authentication middlewares: ](https://docs.djangoproject.com/en/4.0/ref/middleware/#module-django.contrib.auth.middleware)

- Add the middleware
- Write tests for it

Samy-33:
closes #97


Aksh-Bansal-dev (approved):
LGTM.

PR #133, opened 2022-07-09
feat(ui): added logout feature

### LINKED ISSUE #128: [UI] Logout feature
Add a logout function to remove the token from sessionStorage and redirect the user to the login page.

prajjwalkapoor:
closes #128


ui/src/components/Sidebar/Sidebar.tsx
```diff
@@ -15,6 +15,10 @@ import {SidebarRow} from './SidebarRow';
 import styles from './styles.module.css';
 
 export const Sidebar: React.FC<{}> = () => {
+  const logoutHandler = () => {
+    window.sessionStorage.removeItem('token');
+    window.location.href = "/login"
```
Aksh-Bansal-dev, line 20: Try using `useNavigate` hook from react-router-dom. That way we can prevent page reload.

Aksh-Bansal-dev (changes requested):
I would recommend you to setup ESlint and prettier in your editor. 
You can also run `npm run lint:fix` or `npm run format` to fix these linting errors.

Aksh-Bansal-dev (approved):
LGTM.

ui/src/components/Sidebar/Sidebar.tsx
```diff
@@ -11,10 +11,15 @@ import {
...
 } from '../../assets';
 import {SidebarRow} from './SidebarRow';
-
+import {useNavigate} from 'react-router-dom';
 import styles from './styles.module.css';
 
 export const Sidebar: React.FC<{}> = () => {
+  const navigate = useNavigate();
+  const logoutHandler = () => {
+    window.sessionStorage.removeItem('token');
```
Samy-33, line 20: Logout is a multi step process. Following should be done in order:
1. When user clicks on logout, a `DELETE` request should be sent to the API at `/auth/token:revoke`
2. While the request is being processed, the loading should be shown and the whole screen should be disabled.
3. Once the request succeeds, remove the token from `sessionStorage` and navigate user to `/login`

ui/src/components/Sidebar/Sidebar.tsx
```diff
@@ -11,10 +11,15 @@ import {
...
+    navigate('/login', {replace: true});
```
Samy-33, line 21: We can omit `replace` option.

ui/src/components/Sidebar/SidebarRow.tsx
```diff
@@ -4,12 +4,13 @@ import styles from './styles.module.css';
...
   icon: string;
   text: string;
+  onClick?: () => void;
   //   isActive: boolean;
 }
 
 export const SidebarRow: React.FC<Props> = props => {
   return (
-    <div className={styles.container_row}>
+    <div className={styles.container_row} onClick={props.onClick}>
```
Samy-33, line 13: This will throw an error when onClick is not provided by the parent.

Something like this should make sure that we don't call `onClick` when it's null.

```tsx
<div className={styles.container_row} onClick = {() => props.onClick && props.onClick()}>
```

Samy-33 (changes requested):
Add tests for this.

Samy-33:
@prajjwalkapoor, please comment on #128 so that it can be assigned to you.

PR #134, opened 2022-07-09
feat: add basic profile page

Aksh-Bansal-dev:
I have added a basic profile page. We can change the fields and design later.

### DIFF
--- a/ui/src/index.css
+++ b/ui/src/index.css
@@ -23,4 +23,5 @@ body {
   --secondary-color: white;
 
   --off-white: #f6f7fe;
+  --input-bg: #bcd5eb36;
 }
--- a/ui/src/pages/Profile/index.tsx
+++ b/ui/src/pages/Profile/index.tsx
@@ -1,0 +1,18 @@
+import {ProfileCard} from '../../components/ProfileCard';
+import {Sidebar} from '../../components/Sidebar/Sidebar';
+import {TopBar} from '../../components/TopBar';
+import styles from './styles.module.css';
+
+export const Profile: React.FC = () => {
+  return (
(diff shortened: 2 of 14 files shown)


ui/src/pages/Profile/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import styles from './styles.module.css';
+
+export const Profile: React.FC = () => {
+  return (
+    <div className={styles.container}>
+      <div className={styles.NavContainer}>
+        <Sidebar />
+      </div>
+      <div className={styles.mainContent}>
+        <TopBar />
```
Samy-33, line 13: These are duplicated (same in the `Home` component I believe).

It doesn't have to be in this PR, create a separate issue to keep the common part in a common component (maybe `Layout.tsx`)

Samy-33 (changes requested):

ui/src/index.css
```diff
@@ -23,4 +23,5 @@ body {
   --secondary-color: white;
 
   --off-white: #f6f7fe;
+  --input-bg: #bcd5eb60;
```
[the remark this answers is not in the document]
Aksh-Bansal-dev: > This looks like invalid hex for color?

I have decreased the opacity, so it has more characters.

> And also currently the input field colors is too dark. Let's make it lighter.

That we can change once we have a final design.

ui/src/pages/Profile/index.tsx
Aksh-Bansal-dev: Created an issue for this.

Samy-33 (approved):

Issue #116, opened 2022-07-04
Dockerize full application


brijsiyag:
### Is your proposal related to a problem?

No.

### For development - 
- [ ] Dockerfile.dev for UI
- [ ] Create docker-compose.yml
- [ ] Using nginx
- [ ] Using SMTP4DEV or other smtp email server for mailing services in development.
- [ ] A PostgreSQL server for development also( Yet to discuss ).

### For production -
- [ ] Dockerfile with use of nginx for serving build files.
- [ ] Dockerfile running gunicorn for Django Application.
- [ ] PostgreSQL or other db service provider.

brijsiyag:
Will we use a single container for production, or multiple?

Samy-33:
> Will we use a single container for production, or multiple?

Keep two containers. One for the application (UI and API) another for the DB.

API will store staticfiles (uploaded images etc) in a volume attached to the application container.
DB will store the data in a volume attached to the DB container.

PR #139, opened 2022-07-09
build(app): Dockerize full app for development

brijsiyag:
### For development - 
- [x] Dockerfile.dev for UI
- [x] Create docker-compose.yml
- [ ] Using nginx
- [ ] Using SMTP4DEV or other smtp email server for mailing services in development.
- [ ] A PostgreSQL server for development also( Yet to discuss ).
- [ ] Make GitHub Actions to use containers for CI/CT.

### For production -
- [ ] Dockerfile with use of nginx for serving build files.
- [ ] Dockerfile running gunicorn for Django Application.
- [ ] PostgreSQL or other db service provider.

### DIFF
--- a/api/Dockerfile.dev
+++ b/api/Dockerfile.dev
@@ -1,18 +1,18 @@
 # base image  
-FROM python:3.8   
+FROM python:3.8
 
 # setup environment variable for work directory  
-ENV API_HOME=/home/app/webapp  
+ENV API_HOME=/home/app/api
 
 # make work directory  
-RUN mkdir -p $API_HOME  
+RUN mkdir -p $API_HOME
 
 # set work directory  
-WORKDIR $API_HOME  
+WORKDIR $API_HOME
 
 # set environment variables  
 ENV PYTHONDONTWRITEBYTECODE 1
-ENV PYTHONUNBUFFERED 1  
+ENV PYTHONUNBUFFERED 1
 
 # copy requirements file
 COPY requirements.txt $API_HOME
@@ -21,13 +21,19 @@ COPY requirements.txt $API_HOME
 RUN pip install --upgrade pip && pip install -r requirements.txt
 
 # copy api directory to docker's work directory. 
-COPY . $API_HOME  
+COPY . $API_HOME
 
 # make migrations
-RUN chmod +x scripts/makemigrations.sh && ./scripts/makemigrations.sh && python manage.py migrate
+RUN chmod +x scripts/makemigrations.sh && ./scripts/makemigrations.sh
+
+# Migrate all migrations
+RUN python manage.py migrate
+
+# Populate Database with dummy data
+RUN chmod +x scripts/populate_db.sh && scripts/populate_db.sh
 
 # port where the Django app runs  
-EXPOSE 8000  
+EXPOSE 8000
 
 # start server  
-ENTRYPOINT ["python", "manage.py", "runserver", "0.0.0.0:8000"]
+CMD ["python", "manage.py", "runserver", "0.0.0.0:8000"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: "3.9"
+services:
+  api:
+    build:
+      context: ./api
+      dockerfile: Dockerfile.dev
+    volumes:
+      - ./api:/home/app/api
+    ports:
+      - 8000:8000
+    command: python manage.py runserver 0.0.0.0:8000
+  ui:
+    build:
+      context: ./ui
+      dockerfile: Dockerfile.dev
+    volumes:
+      - ./ui:/home/app/ui
+      - /home/app/ui/node_modules
+    ports:
+      - 3000:3000
+    command: npm run dev -- --host
--- a/ui/Dockerfile.dev
+++ b/ui/Dockerfile.dev
@@ -1,0 +1,24 @@
+FROM node:alpine3.16
+
+# Set working DIE
+WORKDIR /home/app/ui
+
+# Upgrade Alpine Packages
+RUN apk -U upgrade
+
+# Add bash to run tests and other scripts
+RUN apk add --no-cache bash
+
+# Copy package.json and package-lock.json
+COPY package*.json .
+
+# Install node packages
+RUN npm ci
+
+# Copy rest of code
+COPY . .
+
+# PORT where UI will run in browser
+EXPOSE 3000
+
+CMD ["npm","run","dev","--","--host"]
--- a/api/scripts/populate_db.sh
+++ b/api/scripts/populate_db.sh
@@ -1,0 +1,1 @@
+python manage.py loaddata user_data


api/Dockerfile.dev
```diff
@@ -21,13 +21,16 @@ COPY requirements.txt $API_HOME
...
+# Migrate all migrations
+RUN python manage.py migrate
 
 # port where the Django app runs  
-EXPOSE 8000  
+EXPOSE 8000
 
 # start server  
-ENTRYPOINT ["python", "manage.py", "runserver", "0.0.0.0:8000"]
+CMD ["python", "manage.py", "runserver", "0.0.0.0:8000"]
```
Samy-33, line 36: Why CMD over ENTRYPOINT?

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.9"
```
Samy-33, line 1: What does 3.9 as a version signify?

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+services:
+  api:
+    build:
+      context: ./api
+      dockerfile: Dockerfile.dev
+    volumes:
+      - ./api:/home/app/api
```
Samy-33, line 8: I am a bit confused about how volumes work here.

In the `Dockerfile.dev` we are `COPY`ing the whole `api` directory to `api` directory of the container. And then we are making it a volume. Won't those two conflict?

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    ports:
+      - 8000:8000
+    command: python manage.py runserver 0.0.0.0:8000
```
Samy-33, line 11: We have a similar command in the `Dockerfile.dev`, what's the purpose of this here?

ui/Dockerfile.dev
```diff
@@ -0,0 +1,24 @@
...
+# Install node packages
+RUN npm install
+
+# Copy rest of code
+COPY . .
+
+# PORT where UI will run in browser
+EXPOSE 3000
+
+CMD ["npm","run","dev","--","--host"]
```
Samy-33, line 24: I would prefer `ENTRYPOINT` over `CMD` for the primary command that the container runs.

Samy-33 (changes requested):

api/Dockerfile.dev
brijsiyag: Because ENTRYPOINT is the command that will always run when container starts and we can't overwrite it.
The simple case can be when someone don't want to run default command and want to start app opening bash first.
If someone doesn't overwrite the CMD command that will work as ENTRYPOINT.
Let me know if it is okay or we should migrate to ENTRYPOINT again.

docker-compose.yml
brijsiyag: This is an optional property, but good practice to write docker-compose version.
Should I remove this?

Samy-33: Is it the `docker-compose` version or our file's/app's version?

api/Dockerfile.dev
Samy-33: Hmm, goal of this `Dockerfile.dev` is to allow users to `run` the api or ui server. I can't think of a reason why we should allow overriding of this command.

If there is a case where it will be helpful to the developer then we should use CMD. For now, I can't think of any.

docker-compose.yml
brijsiyag: This is the command for docker-compose only, in future if we want to change the CMD in Dockerfile for running only if a single container started without docker-compose, than this command will overwrite Dockerfile's command and used only in docker-compose.

Samy-33:
This PR doesn't fix #116 completely, so you can remove the link from the description.

api/Dockerfile.dev
brijsiyag: As I mentioned above the case might be if we want to start server on our own using `bash`.
We can use ENTRYPOINT in prod Dockerfile.

docker-compose.yml
brijsiyag: It is the `docker-compose` version.

Samy-33: Let's keep it then.

docker-compose.yml
brijsiyag: No, making volume is just mapping both directories so that if any files changed at place the changes will also reflect at the mapped directory.

Samy-33: Okay. Suppose you copy some other files to containers `api` directory. And then map your `api` directory to container's `api` directory. What happens in that case?

The files are different now.

Aksh-Bansal-dev:
Do we have to specify the path for `Dockerfile.dev` in the `docker build` command if we only want to build the API image?
If yes, please make changes in the `api` docs for this.

ui/Dockerfile.dev
```diff
@@ -0,0 +1,24 @@
...
+RUN apk -U upgrade
+
+# Add bash to run tests and other scripts
+RUN apk add --no-cache bash
+
+# Copy package.json and package-lock.json
+COPY package*.json .
+
+# Install node packages
+RUN npm install
```
Aksh-Bansal-dev, line 16: This should be `npm ci` as we don't want to have different npm package version than what we have in package-lock.json.

Aksh-Bansal-dev, line 16: This should be `npm ci` as we don't want to have different npm package version than what we have in package-lock.json.

api/Dockerfile.dev
Aksh-Bansal-dev: According to [Docker docs](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact):
> ENTRYPOINT should be defined when using the container as an executable.

In our case, we are using it as an executable to run API or UI server.
Also, we can override the entrypoint using `--entrypoint` flag.

docker-compose.yml
brijsiyag: > In general, the Docker build command restricts the sources of files we can use in our Docker images. We
specify a build context, which is the root from which both the Dockerfile and all its dependent files must be
found.

I don't think we will be copying files from outside of the root of Dockerfile.

api/Dockerfile.dev
brijsiyag: My view is the ENTRYPOINT is more suitable for production and for development setup we should use CMD because as I debug containers I don't use CMD or ENTRYPOINT and open bash to start the server.
If this is not the case than I will change CMD to ENTRYPOINT again.

brijsiyag:
> Do we have to specify the path for `Dockerfile.dev` in the `docker build` command if we only want to build the API image?
> If yes, please make changes in the `api` docs for this.

Yes, but Dockerfile.dev is not meant to be used separately. It will used with docker-compose so we have to update docs with full setup using docker.
I will create a new issue to update docs.

docker-compose.yml
Samy-33: I just want to understand what happens in that case. If it results in a conflict with different files, then the way we are using volume rn is not the way to do it.

brijsiyag: To simulate this I tried to build a image with a outer reference , got following error - 
```
=> ERROR [4/4] COPY /Users/brijsiyag/Desktop/docker-k8s/docker-django/manage.py .                                                               0.0s
------
 > [4/4] COPY /Users/brijsiyag/Desktop/docker-k8s/docker-django/manage.py .:
------
failed to compute cache key: "/Users/brijsiyag/Desktop/docker-k8s/docker-django/manage.py" not found: not found
```

Samy-33: Cool, looks good.

api/Dockerfile.dev
Samy-33: Cool, I am okay with using CMD in dev env. Make sure to use ENTRYPOINT for prod.

brijsiyag: Sure, Thank you !!

brijsiyag:
```
raise InconsistentMigrationHistory(
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration admin.0001_initial is applied before its dependency api_accounts.0001_initial on database 'default'.
```
I am encountering this error while migrating.
Should we make a script to ensure migrations happen in correct order.
and we can migrate database to PostgreSQL with docker setup as it would be easy to debug database.

brijsiyag:
@Samy-33 , This PR is ready for review.  PTAL

Samy-33:
Running `docker run gymkhana_api` does not accept `-p` as a parameter

Aksh-Bansal-dev:
I usually use `docker run --name gymkhana-api -p 8000:8000 gymkhana-api` to run api server.

[reference](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose).

Samy-33:
Works, thanks @Aksh-Bansal-dev and @brijsiyag

Samy-33 (approved):

PR #144, opened 2022-07-10
fix(api): delete existing token and return new token

PranshuNayak:


api/api/auth/tests.py
```diff
@@ -99,7 +99,7 @@ def test_createToken_userExists_tokenDoesNotExist_createsTokenAndReturnsWithUser
     @mock.patch('time.sleep')
     @mock.patch('google_auth_oauthlib.flow.Flow.fetch_token')
     @mock.patch('google.oauth2.id_token.verify_oauth2_token')
-    def test_createToken_tokenExists_returnsExistingTokenWithUserInfo(
+    def test_createNewToken_tokenExists_returnsNewTokenWithUserInfo(
```
Samy-33, line 102: Try to not edit existing tests if new they are needed. Add a new test for your funcionality.

Samy-33 (changes requested):

api/api/auth/tests.py
PranshuNayak: > Try to not edit existing tests if new they are needed. Add a new test for your functionality.

This tests gets its response from `/auth/token` and the `post` method in `TokenView` uses different function which return a `new token` so this test will always fail

PranshuNayak: > You modified the `find_or_create` method. Which now always returns a new token regardless of whether it's expired or not.
> 
> The test would pass if the `find_or_create` method stays the same. Figure out a way to remove the expired token and create a new one only if it's expired.
> 
> No need to generate a new token every time TokenView is called.

Probably I am confused with the implementation. I did not see any criteria to validate whether the token is expired or not .I think we need to set some?

Samy-33: There's a `created_at` field in the `Token` model. Use that to check if it's expired.

Token is expired after 1 day.

PranshuNayak: > There's a `created_at` field in the `Token` model. Use that to check if it's expired.
> 
> Token is expired after 1 day.

Sir we need to modify  this `def test_createToken_tokenExists_returnsExistingTokenWithUserInfo` coz it does not satisfy our way of implementation. It should be `def test_createToken_validTokenExists_returnsExistingTokenWithUserInfo`

PranshuNayak:
@Samy-33 made the changes

api/api/auth/tests.py
```diff
@@ -138,6 +139,51 @@ def test_createToken_tokenExists_returnsExistingTokenWithUserInfo(
                          f'Got {users_count} users in db, expected 1')
         self.assertEqual(response.data['token'], existing_token.token)
 
+    @pytest.mark.django_db
+    @mock.patch('time.sleep')
+    @mock.patch('google_auth_oauthlib.flow.Flow.fetch_token')
+    @mock.patch('google.oauth2.id_token.verify_oauth2_token')
+    def test_createExpiredToken_tokenExists_returnsNewTokenWithUserInfo(
```
Samy-33, line 146: Test should not describe the implementation detail of the test.
just say, `expiredTokenExists` instead of `createExpiredToken`.
When somebody reads a test, they want to know what the test does instead of how it does it.

api/api/auth/views.py
```diff
@@ -51,7 +52,10 @@ def post(self, request):
 
             user = self._find_or_create_user(user_info)
             access_token = Token.find_or_create(user)
+            token_life = datetime.now(timezone.utc) - access_token.created_at
 
+            if(token_life.days >= 1):
```
Samy-33, line 57: extract 1 in a constant variable. Maybe `TOKEN_LIFE_IN_DAYS`.

That way it would be much more readable and could be used somewhere else if needed.

PranshuNayak:
@Samy-33 made the changes

Samy-33 (changes requested):

api/api/auth/models.py
```diff
@@ -16,3 +17,10 @@ def find_or_create(user):
         if not token:
             return Token.objects.create(token=uuid.uuid4().hex, user=user)
         return token
+
+    def if_expired_get_new(token, user):
+        TOKEN_LIFE_IN_DAYS = (datetime.now(timezone.utc) - token.created_at).days
```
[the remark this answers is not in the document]
PranshuNayak: Can we rename this as `AUTH_TOKEN_LIFE_IN_DAYS` and place it below `AUTH_USER_MODEL` 
Reason : Seemed more consistent in settings.py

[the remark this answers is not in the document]
Samy-33: Sounds good.

PR #155, opened 2022-07-15
feat(api): add endpoint to get all clubs details

PranshuNayak:


api/api/clubs/tests.py
```diff
@@ -0,0 +1,35 @@
...
+
+from api.clubs.models import Club
+
+User = get_user_model()
+
+
+class ListClubsViewTest(APITestCase):
+    """Tests ListClubView for getting the list of clubs."""
+
+    def testListClubs_noClubExistsInDb_returnsEmptyList(self):
```
Samy-33, line 12: mark these tests with `@python.markers.django_db`, it ensures that a clean db is used for each test and cleans up after the test finishes: [reference](https://pytest-django.readthedocs.io/en/latest/helpers.html#pytest.mark.django_db)

Look at tests in other files and refactor them as well.

api/api/clubs/serializers.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from api.clubs.models import Club
+
+
+class ClubsSerializer(serializers.ModelSerializer):
```
Samy-33, line 6: Let's keep it singular, as it's used to represent a single Club.

Samy-33 (changes requested):
Overall good, few small changes.

api/api/clubs/tests.py
PranshuNayak: > mark these tests with `@python.markers.django_db`, it ensures that a clean db is used for each test and cleans up after the test finishes: [reference](https://pytest-django.readthedocs.io/en/latest/helpers.html#pytest.mark.django_db)
> 
> Look at tests in other files and refactor them as well.

Done!

api/api/auth/tests.py
```diff
@@ -159,6 +160,7 @@ def test_createToken_invalidAuthorizationCodeInRequest_throwsForbiddenError(
 
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
+    @pytest.mark.django_db
```
Samy-33, line 163: This test doesn't use a database.

Samy-33 (approved):
LGTM

api/api/auth/tests.py
PranshuNayak: here it does
`tokens_count = Token.objects.count()`
`users_count = User.objects.count()`

PR #164, opened 2022-07-15
feat(api): add endpoint for creating users

arminpatel:
+ `view` : generic CreateAPI view
+ `Post` request at `/users/` for creating users
+ ` Serializers` 
  + Simple Model Serializers for `Faculty`, `Student` and `Staff`
  + These three serializers are nested inside `UserSerializer` and their `required` attribute is set to `False`
  + The `create` method then extracts data of each Serializer(if present) and the remaining data is used to create the user. And depending on the extracted data, the student, faculty and staff instances are created.
  +  The problem I was facing in this implementation was that a User could be only one of staff, faculty and student. I've have handled it by assuming only one or neither of them would be passed to the api. Let me know if the design could have been better.
+ `testing`: I have added some test to cover different scenarioes like adding base User, Student, Faculty etc, and what should happen when invalid emails are provided


api/api/accounts/serializers.py
```diff
@@ -0,0 +1,48 @@
+from rest_framework import serializers
+
+from .models import Faculty, Staff, Student, User
```
Samy-33, line 3: `get_user_model()` should be used instead of `User`

api/api/accounts/views.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import generics
+
+from .models import User
+from .serializers import UserSerializer
+
+
+class UsersListView(generics.CreateAPIView):
```
Samy-33, line 7: Only the AdminUsers should be able to call this API.

We don't want just anyone to create users. On SignUp the users are created. And if needed this API will be used to create Users only be the admin.

Tests will change accordingly.

api/api/accounts/serializers.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class UserSerializer(serializers.ModelSerializer):
+    student = StudentSerializer(many=False, allow_null=True, required=False)
+    faculty = FacultySerializer(many=False, allow_null=True, required=False)
+    staff = StaffSerializer(many=False, allow_null=True, required=False)
+
+    class Meta:
+        model = User
+        fields = ['id', 'email', 'first_name', 'last_name', 'gender',
```
Samy-33, line 31: id should be a read only field. Or it can be removed as the end user doesn't create the id.

api/api/accounts/tests.py
```diff
@@ -0,0 +1,132 @@
...
+    @pytest.mark.django_db
+    def testUsersListView_invalidEmailsProvided_returnsHttp400(self):
+        # given
+        user = User.objects.create(
+            email=primaryTestUserMail, first_name='Test', last_name='User')
+        self.client.force_authenticate(user)
+
+        invalidTestMail1 = primaryTestUserMail
+        invalidTestMail2 = ''
+        invalidTestMail3 = 'testusercom'
```
Samy-33, line 37: You can use [@pytest.mark.parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) to have test data defined separately from the test implementation.

This way test can focus on the actual logic for testing.

api/api/accounts/tests.py
```diff
@@ -0,0 +1,132 @@
...
+
+User = get_user_model()
+userCreationEndpoint = '/users/'
+primaryTestUserMail = 'test@user.com'
+
+
+class UsersListViewTest(APITestCase):
+
+    @pytest.mark.django_db
+    def testUsersListView_testBaseUserModel_validEmailProvided_userSuccessfullyCreated(self):
```
Samy-33, line 14: All these test names should be updated as it's not `UsersListView`

arminpatel:
Sorry closing this and creating new pr

api/api/accounts/tests.py
```diff
@@ -0,0 +1,132 @@
+import pytest
+from django.contrib.auth import get_user_model
+from rest_framework.test import APITestCase
+from rest_framework import status
+
+User = get_user_model()
+userCreationEndpoint = '/users/'
```
Samy-33, line 7: variables should follow snake_case: https://peps.python.org/pep-0008/#function-and-variable-names

api/api/accounts/tests.py
```diff
@@ -0,0 +1,132 @@
...
+                                      'staff': {
+                                          'department': 'ECE',
+                                          'designation': 'z'
+                                      }
+                                    })
+
+        # then
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data['email'], testUserEmail)
+        self.assertEqual(response.data['staff']['department'], 'ECE')
```
Samy-33, line 132: Let's assert for the whole user object.

Samy-33 (changes requested):

PR #166, opened 2022-07-15
feat(api): add endpoint for creating users

arminpatel:
+ `view` : generic CreateAPI view
+ `Post` request at `/users/` for creating users
+ ` Serializers` 
  + Simple Model Serializers for `Faculty`, `Student` and `Staff`
  + These three serializers are nested inside `UserSerializer` and their `required` attribute is set to `False`
  + The `create` method then extracts data of each Serializer(if present) and the remaining data is used to create the user. And depending on the extracted data, the student, faculty and staff instances are created.
  +  The problem I was facing in this implementation was that a User could be only one of staff, faculty and student. I've have handled it by assuming only one or neither of them would be passed to the api. Let me know if the design could have been better.
+ `testing`: I have added some test to cover different scenarios like adding base User, Student, Faculty etc, and what should happen when invalid emails are provided

### DIFF
--- a/api/api/accounts/tests.py
+++ b/api/api/accounts/tests.py
@@ -1,0 +1,132 @@
+import pytest
+from django.contrib.auth import get_user_model
+from rest_framework.test import APITestCase
+from rest_framework import status
+
+User = get_user_model()
+userCreationEndpoint = '/users/'
+primaryTestUserMail = 'test@user.com'
+
+
+class UsersListViewTest(APITestCase):
+
+    @pytest.mark.django_db
+    def testUsersListView_testBaseUserModel_validEmailProvided_userSuccessfullyCreated(self):
+        # given
+        user = User.objects.create(
+            email=primaryTestUserMail, first_name='Test', last_name='User')
+        self.client.force_authenticate(user)
+        testUserEmail = 'test1@user.com'
+
+        # when
+        response = self.client.post(userCreationEndpoint, {'email': testUserEmail}, format='json')
+
+        # then
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data['email'], testUserEmail)
+
+    @pytest.mark.django_db
+    def testUsersListView_invalidEmailsProvided_returnsHttp400(self):
+        # given
+        user = User.objects.create(
+            email=primaryTestUserMail, first_name='Test', last_name='User')
+        self.client.force_authenticate(user)
+
+        invalidTestMail1 = primaryTestUserMail
+        invalidTestMail2 = ''
+        invalidTestMail3 = 'testusercom'
+
+        # when
+        response1 = self.client.post(userCreationEndpoint, {'email': invalidTestMail1},
+                                     format='json')
+        response2 = self.client.post(userCreationEndpoint, {'email': invalidTestMail2},
+                                     format='json')
+        response3 = self.client.post(userCreationEndpoint, {'email': invalidTestMail3},
+                                     format='json')
+
+        # then
+        self.assertEqual(response1.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response2.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response3.status_code, status.HTTP_400_BAD_REQUEST)
+
+    @pytest.mark.django_db
+    def testUsersListView_testStudentModel_validDataProvided_userSuccessfullyCreated(self):
+        # given
+        user = User.objects.create(
+            email='test@user.com', first_name='Test', last_name='User')
+        self.client.force_authenticate(user)
+
+        testUserEmail = 'test@student.com'
+
+        # when
+        response = self.client.post(userCreationEndpoint,
+                                    {
+                                      'email': testUserEmail,
+                                      'first_name': "test",
+                                      'user_type': 'Student',
+                                      'student': {
+                                          'roll_no': "21abc000",
+                                          'batch': 2025,
+                                          'department': "CSE",
+                                          'hostel_address': "H3-Z000"
+                                      }
+                                    })
+
+        # then
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data['email'], testUserEmail)
+        self.assertEqual(response.data['student']['roll_no'], '21abc000')
+
+    @pytest.mark.django_db
+    def testUserListView_testFacultyModel_validDataProvided_userSuccessfullyCreated(self):
+        # given
+        user = User.objects.create(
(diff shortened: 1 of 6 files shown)


Samy-33:
Left some comments on https://github.com/BitByte-TPC/gymkhana/pull/164.

Closing a PR if you mess something up in git and reopening a fresh PR is not good for multiple reasons:
1. You have to do this toiling task every time you mess up something in `git`.
2. Reviews done in the previous PR is kind of lost.
3. Git provides ways to do it in an easier way, so you should utilise that. 
    - [git reflog](https://www.atlassian.com/git/tutorials/rewriting-history/git-reflog#:~:text=Git%20keeps%20track%20of%20updates,git%20reset)
    - [git rebase](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase)

arminpatel:
> Left some comments on #164.
> 
> Closing a PR if you mess something up in git and reopening a fresh PR is not good for multiple reasons:
> 
> 1. You have to do this toiling task every time you mess up something in `git`.
> 2. Reviews done in the previous PR is kind of lost.
> 3. Git provides easier ways to fix the mess up, so you should utilise that.
>    
>    * [git reflog](https://www.atlassian.com/git/tutorials/rewriting-history/git-reflog#:~:text=Git%20keeps%20track%20of%20updates,git%20reset)
>    * [git rebase](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase)

Ok, will mind from next time. So should I fix and reopen #164 or keep this one?

arminpatel:
Tried to use pytest parametrize and fixtures but was thrown with the following error,

Screenshot 2022-07-16 at 8 28 13 AM

[my code](https://p.ip.fi/b6Ih)

Pytest fixtures and parametrize do not support django unittest.TestCase subclass
[reference](https://docs.pytest.org/en/latest/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses)
So is there any other way?

Samy-33:
Have you tried placing `@parameterize` decorator below the `@django_db` decorator?

As decorator's order matters in python, it might be the issue.

- Don't use variables in the parameters list, just use the values. Readers have to go to the variable definition to know the value.

-`invalidEmailProvided`, `throwsBadRequest`

arminpatel:
> Have you tried placing `@parameterize` decorator below the `@django_db` decorator?
> 
> As decorator's order matters in python, it might be the issue.

Still getting the same error. [my code](https://p.ip.fi/JTks)

Samy-33:
@arminpatel can you push your code, I will take a look.

Samy-33:
@arminpatel let's use function based tests for `CreateUserView`

```python
@pytest.mark.parametrize('invalid_email', ['', 'someemail'])
@pytest.mark.django_db
def testUsersListView_invalidEmailsProvided_returnsHttp400(invalid_email):
    # given
    user = User.objects.create(
        email=primaryTestUserMail, first_name='Test', last_name='User')
    client = APIClient()
    client.force_authenticate(user)

    # when
    response1 = client.post(userCreationEndpoint, {'email': invalid_email},
                            format='json')

    # then
    assert response1.status_code == status.HTTP_400_BAD_REQUEST
```

arminpatel:
> @arminpatel let's use function based tests for `CreateUserView`
> We need to refactor all the tests to be in functional format instead of class based.

ok, I will make the changes

arminpatel:
@Samy-33 Made the changes.

api/api/accounts/tests.py
```diff
@@ -0,0 +1,170 @@
...
+                            'user_type': 'Staff',
+                            'staff': {
+                                'department': 'ECE',
+                                'designation': 'z'
+                                }
+                            })
+
+    # then
+    assert response.status_code == status.HTTP_201_CREATED
+    assert response.data['email'] == test_user_email
```
Samy-33, line 158: redundant check?

api/api/accounts/tests.py
```diff
@@ -0,0 +1,170 @@
...
+                             'user_type': 'Student',
+                             'picture_url': '',
+                             'student': None,
+                             'faculty': None,
+                             'staff': None
+                             }
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('invalid_mail', ['test@user.com', '', 'testusersom'])
```
Samy-33, line 60: Let's add comment for each invalid email as to why it's invalid.

```python
@pytest.mark.parametrize('invalid_mail', [
    'test@user.com', # admin user 
    '', # empty email
    'testusersom' # invalid email format
])
```

api/api/accounts/views.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import generics
+from rest_framework.permissions import IsAdminUser
+
+from .models import User
```
Samy-33, line 4: use `get_user_model()`

api/api/accounts/tests.py
```diff
@@ -0,0 +1,170 @@
...
+user_creation_endpoint = '/users/'
+primary_test_user_mail = 'test@user.com'
+
+
+@pytest.mark.django_db
+def testCreateUserView_testPermission_nonAdminRequest_throwsForbidden():
+    # given
+    user = User.objects.create(
+        email=primary_test_user_mail, first_name='Test', last_name='User')
+    client = APIClient()
```
Samy-33, line 18: nit: we can use `pytest.fixture` to create repetitive objects once.

```python
@pytest.mark.django_db
@pytest.fixture(scope='function')
def admin_user(django_user_model):
    return django_user_model.objects.create(email=primary_test_user_mail, first_name='Test', last_name='User', is_staff=True)

@pytest.fixture(scope='function')
def client():
    return APIClient()
```

And then use it like:
```python
@pytest.mark.django_db
def testCreateUserView_testPermission_nonAdminRequest_throwsForbidden(admin_user, client):
    ...
```

`pytest_django` also allows you to use `django_user_model` fixture to get `User` model: [Reference](https://pytest-django.readthedocs.io/en/latest/helpers.html#django-user-model)

api/api/accounts/tests.py
```diff
@@ -0,0 +1,170 @@
...
+                            'student': {
+                                'roll_no': "21abc000",
+                                'batch': 2025,
+                                'department': "CSE",
+                                'hostel_address': "H3-Z000"
+                                }
+                            })
+
+    # then
+    print(response.data)
```
Samy-33, line 100: Remove print

Samy-33 (changes requested):
Overall looks good, few minor improvements.

api/api/accounts/tests.py
arminpatel: Yeah, thought more checks wouldn't hurt 😅 . I will remove them.

Samy-33 (approved):
Looks great. Good work @arminpatel

PR #170, opened 2022-07-16
feat(api): add pagination for list apis

PranshuNayak:


Samy-33:
Can you figure out a way to always cap the max results to 50?

PranshuNayak:
Created a custom pagination class with `LimitOffsetPagination` and set it as `default` however unable to test it 
c
.

PR #172, opened 2022-07-16
feat(api): add a delete handler for tokenview

PranshuNayak:


api/api/auth/urls.py
```diff
@@ -1,7 +1,8 @@
 from django.urls import path
 
-from .views import TokenView
+from .views import DeleteTokenView, TokenView
 
 urlpatterns = [
     path('token/', TokenView.as_view()),
+    path('token:token/', DeleteTokenView.as_view())
```
Samy-33, line 7: should be `token:revoke`

api/api/auth/views.py
```diff
@@ -71,3 +73,20 @@ def _find_or_create_user(self, user_info):
                 first_name=user_info.get('first_name'),
                 last_name=user_info.get('last_name'))
         return user
+
+
+class DeleteTokenView(APIView):
+    """View to delete the tokens."""
+
+    def delete(self, request):
+        serializer = DeleteTokenRequestSerializer(data=request.data)
```
Samy-33, line 82: Now that I think about it, we don't even need a `token` in the request `body`. If the user is authenticated, you can access the `user` from `request.user`.

Then use `related_name` to access it's token.

```python
user_token = request.user.token
if token:
    token.delete()
```

We dont need to throw an error if token doesn't exist. We can simply consider it deleted and return `HTTP_204`.

api/api/auth/views.py
```diff
@@ -71,3 +73,20 @@ def _find_or_create_user(self, user_info):
...
+        if serializer.is_valid():
+            token = serializer.validated_data.get('token')
+            try:
+                existing_token = Token.objects.get(token=token)
+                existing_token.delete()
+                return Response(status=status.HTTP_204_NO_CONTENT)
+            except ObjectDoesNotExist:
+                return Response(status=status.HTTP_403_FORBIDDEN,
+                                data={'message': 'Token could not be validated'})
```
Samy-33, line 91: Not required. If token doesn't exist, consider it deleted.

Samy-33 (changes requested):

api/api/auth/views.py
PranshuNayak: > 

If the user is authenticated then the token will always be valid, isn't it?

Samy-33: Think about this case: User logs in using two devices.

PranshuNayak: Ok so from our model implementation of `Token` a user can have multiple `tokens` . When we use `request.user.token` to `delete` the token how would we know which of the tokens to be deleted? Or we are aiming to logout the `user` from all the devices.

`Why a user can have multiple tokens ?`
AFAIK since we haven't specified `primary key=true` on `user` attribute in `Token model` due to which `django` will treat the `id` column as `primary key` so a `user` can have multiple tokens

Samy-33: `user` is a `OneToOneField` in `Token` model:
https://github.com/BitByte-TPC/gymkhana/blob/4fb305f8ac76daf21b7321618b3f233704414f9e/api/api/auth/models.py#L10

Samy-33 (approved):
LGTM

PR #173, opened 2022-07-16
feat(api): endpoint for updating users

arminpatel:


api/api/accounts/serializers.py
```diff
@@ -49,3 +49,84 @@ def create(self, validated_data):
...
+
+    def _update_base_user(self, instance, validated_data):
+        instance.email = validated_data.get('email', instance.email)
+        instance.first_name = validated_data.get('first_name', instance.first_name)
+        instance.last_name = validated_data.get('last_name', instance.last_name)
+        instance.gender = validated_data.get('gender', instance.gender)
+        instance.contact_no = validated_data.get('contact_no', instance.contact_no)
+        instance.user_type = validated_data.get('user_type', instance.user_type)
+        instance.is_active = validated_data.get('is_active', instance.is_active)
+        instance.picture_url = validated_data.get('picture_url', instance.picture_url)
```
Samy-33, line 98: Can we do something like:

```python
updatable_fields = ['email', 'first_name', 'last_name', ...]

for field in updatable_fields:
  field_stored_value = getattr(instance, field)
  field_new_value = validated_data.get(field, field_stored_value)
  setattr(instance, field, field_new_value)
```

Maybe extract this piece in a helper method:
```python
def _update_instance(self, instance, validated_data, updatable_fields):
  ...
```

api/api/accounts/test_update_users.py
```diff
@@ -0,0 +1,307 @@
...
+
+    # when
+    response = client.put(f'/users/{new_user.id}/', {'email': 'changed_mail@user.com'})
+
+    # then
+    assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+@pytest.mark.django_db
+def testUpdateUserView_baseUser_ownerChangeRequest_updatesSuccessful(client, test_user):
```
Samy-33, line 44: `baseUser` -> `onlyBaseUserInRequest`

api/api/accounts/serializers.py
```diff
@@ -49,3 +49,84 @@ def create(self, validated_data):
...
+            Faculty.objects.create(user=instance, **faculty_data)
+        elif staff:
+            self._update_staff(staff, staff_data)
+        elif staff_data:
+            Staff.objects.create(user=instance, **staff_data)
+
+        return instance
+
+    def _update_base_user(self, instance, validated_data):
+        instance.email = validated_data.get('email', instance.email)
```
Samy-33, line 91: `email`, `user_type` and `student.roll_no` should not be allowed to be changed, unless update is being done by an `admin` user. See if any other such field need to be considered.

Also, make sure that a user of one type does not add info about other type, like `user_type is student` and a `faculty` field is being updated.

api/api/accounts/test_update_users.py
```diff
@@ -0,0 +1,307 @@
...
+
+    client.force_authenticate(admin_user)
+
+    # when
+    response = client.put(f'/users/{test_user.id}/', {'email': 'test@user.com',
+                                                      'first_name': 'changed_name'})
+    test_user = User.objects.get(id=test_user.id)
+
+    # then
+    print(test_user)
```
Samy-33, line 76: remove print.

api/api/accounts/test_update_users.py
```diff
@@ -0,0 +1,307 @@
...
+    admin_user = User.objects.create(email='admin@user.com',
+                                     first_name='admin',
+                                     is_staff=True)
+
+    client.force_authenticate(admin_user)
+
+    # when
+    response = client.put(f'/users/{test_user.id}/', {'email': 'test@user.com',
+                                                      'first_name': 'changed_name'})
+    test_user = User.objects.get(id=test_user.id)
```
Samy-33, line 73: I believe this should move to `# then` section as it's a result of calling response

Should be called `updated_test_user`

api/api/accounts/test_update_users.py
```diff
@@ -0,0 +1,307 @@
...
+
+    assert test_user.faculty.designation == 'smthng_different'
+    assert test_user.gender == 'F'
+
+
+@pytest.mark.django_db
+def testUpdateUserView_staffUser_changeRequest_updatesSuccessful(client, test_user):
+
+    # given
+    staff = Staff.objects.create(user=test_user,                # noqa: F841
```
Samy-33, line 277: Try not to use comments to skip lint checks. They should be added only when there's no way to avoid them.

In this case, you can remove the assignment: `staff = `

Samy-33 (changes requested):

api/api/accounts/test_update_users.py
arminpatel: I am going to write a python script that does this for me 😆 .

Samy-33: hmmm. I have added `flake8-print` to catch that in CI and your local precommit if you've configured them in this [commit](https://github.com/BitByte-TPC/gymkhana/commit/7a9ebf972ed498157facf7192465f536e5f05c4b).

You can run `flake8 api` from your api directory, it should show all the flake8 errors along with any print statements.

arminpatel: That's great. Thanks.

api/api/accounts/serializers.py
arminpatel: This is much better.

api/api/accounts/serializers.py
```diff
@@ -25,27 +26,159 @@ class Meta:
...
+                Faculty.objects.create(user=user, **faculty_data)
+            else:
+                raise serializers.ValidationError('Provided data doesn\'t match with user_type')
         elif staff_data:
-            Staff.objects.create(user=user, **staff_data)
-
+            if user_type == 'Staff':
+                Staff.objects.create(user=user, **staff_data)
+            else:
+                raise serializers.ValidationError('Provided data doesn\'t match with user_type')
```
Samy-33, line 66: Instead of throwing and error why don't we just use the `student`, `staff` or `faculty` fields according to the `user_type`?

I think we should avoid throwing errors unless they are essential and make user's life easier/better.
[Reference](https://youtu.be/bmSAYlu0NcY?t=1693)

api/api/accounts/serializers.py
```diff
@@ -25,27 +26,159 @@ class Meta:
...
+                                                      ):
+        request_user = self.context['request'].user
+
+        if not request_user.is_staff:
+            for field in admin_only_editable_fields:
+                field_stored_value = getattr(instance, field)
+                field_new_value = validated_data.get(field, None)
+
+                if field_new_value and field_stored_value != field_new_value:
+                    raise serializers.ValidationError(f'{field} field is not editable')
```
Samy-33, line 180: Again instead of throwing this error, let's just ignore those fields?

Avoid exceptions when they can be.

api/api/accounts/test_update_users.py
```diff
@@ -0,0 +1,455 @@
...
+                                            last_name='user')
+
+
+@pytest.fixture(scope='function')
+def client():
+    return APIClient()
+
+
+@pytest.mark.django_db
+def testUpdateUserView_nonOwnerRequest_throwsForbidden(client, test_user):
```
Samy-33, line 28: `nonOwnerAdNonAdminRequest`

Samy-33 (changes requested):
It might go through 1-2 more iterations. But keep up.

api/api/accounts/serializers.py
```diff
@@ -25,27 +26,159 @@ class Meta:
...
                   'contact_no', 'user_type', 'picture_url', 'student', 'faculty', 'staff']
+        extra_kwargs = {'email': {'required': False}}
 
     def create(self, validated_data):
         student_data = validated_data.pop('student', None)
         faculty_data = validated_data.pop('faculty', None)
         staff_data = validated_data.pop('staff', None)
 
+        # check for email
+        if not validated_data.get('email', None):
```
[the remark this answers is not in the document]
Samy-33: Ah my bad. It's create method, I thought it was `update`.

api/api/accounts/serializers.py
arminpatel: Hmm, that would better for the user. But error should still thrown when user_type and data doesn't match while updating right?

arminpatel: > Instead of throwing and error why don't we just use the `student`, `staff` or `faculty` fields according to the `user_type`?
> Wouldn't it be better if I change user_type according to data provided?

Samy-33: What if the user puts `student` and `faculty` fields? How will you choose the `user_type`?

arminpatel: Yeah, that case couldn't handled. But if user_type is `student` and data provided is of `faculty`, we would just ignore the request?

arminpatel: Maybe a mix all both cases? When multiple types of data is provided take according to user type and when there is one type of data but mismatched user_type change user_type and ignore request in any other cases. Would this add unnecessary complexity? 🤔

Samy-33: It's an overkill, we want to do the simplest thing that solves the problem. [KISS](https://en.wikipedia.org/wiki/KISS_principle).

api/api/accounts/serializers.py
```diff
@@ -25,27 +26,121 @@ class Meta:
...
```
Samy-33, line 45: `dict.get` by default returns None if not found.

api/api/accounts/serializers.py
```diff
@@ -25,27 +26,121 @@ class Meta:
...
+            faculty = None
+
+        try:
+            staff = instance.staff
+        except ObjectDoesNotExist:
+            staff = None
+
+        user_type = instance.user_type
+
+    # create or update user type only if user_type and data match else ignore it
```
Samy-33, line 86: indentation?

api/api/accounts/test_create_users.py
```diff
@@ -176,3 +176,44 @@ def testCreateUserView_testStaffModel_validDataProvided_userCreatedSuccessfully(
...
+                             'gender': 'M',
+                             'contact_no': None,
+                             'user_type': 'Faculty',
+                             'picture_url': '',
+                             'student': None,
+                             'faculty': OrderedDict([('title', 'Prof'),
+                                                     ('department', 'ME'),
+                                                     ('designation', None)]),
+                             'staff': None
+                             }
```
Samy-33, line 219: also assert the db entry

Samy-33 (changes requested):

Samy-33 (approved):
LGTM

PR #177, opened 2022-07-18
feat(api): add pagination class for list apis

PranshuNayak:


Samy-33:
Can you add test for it? Maybe in list clubs tests, you can add a test by creating 100 clubs and checking if first request only returns 50 or not.

api/api/pagination.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework.pagination import LimitOffsetPagination
+
+
+class DefaultPagination(LimitOffsetPagination):
```
Samy-33, line 4: Call this APILimitOffsetPagination

Samy-33 (changes requested):

PranshuNayak:
> Can you add test for it? Maybe in list clubs tests, you can add a test by creating 100 clubs and checking if first request only returns 50 or not.

Done!

Samy-33 (approved):
LGTM

PR #181, opened 2022-07-19
feat(api): add endpoint to update club details

arminpatel:


api/api/clubs/serializers.py
```diff
@@ -6,4 +6,7 @@
 class ClubSerializer(serializers.ModelSerializer):
     class Meta:
         model = Club
-        fields = ['name', 'category', 'logo']
+        fields = ['name', 'category', 'logo', 'description', 'email']
```
PranshuNayak, line 9: update URL uses `<int:pk>` for `updation` . We need to send `Club id` to the Client

api/api/clubs/tests.py
```diff
@@ -55,3 +56,55 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+                                        logo="#")
+        # when
+        response = self.client.put(f'/clubs/{test_club.id}/', {'name': 'changed_club_name',
+                                                               'category': "S&T",
+                                                               'description': "Some desc here",
+                                                               'email': "theprogclub@iiitdmj.ac.in",
+                                                               'logo': "https://iiitdmj.ac.in"})
+
+        # then
+        updated_club = Club.objects.get(id=test_club.id)
```
PranshuNayak, line 106: We can use `pk` instead of `id`

api/api/clubs/tests.py
```diff
@@ -55,3 +56,55 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data['name'], 'changed_club_name')
+        self.assertEqual(updated_club.name, 'changed_club_name')
+        self.assertEqual(updated_club.category, 'S&T')
```
PranshuNayak, line 110: if `serialization.is_valid` fails a `BadRequestError` needs to be thrown . Can add `test` for it

api/api/clubs/tests.py
arminpatel: Does it really matter? 🤔 
I'll change it anyway.

PranshuNayak: Got this. Anything is okay I suppose .
[https://stackoverflow.com/questions/2165865/django-queries-id-vs-pk
](url)

Samy-33 (changes requested):

api/api/clubs/serializers.py
```diff
@@ -6,4 +6,7 @@
...
+        extra_kwargs = {'description': {'write_only': True},
+                        'email': {'write_only': True}
```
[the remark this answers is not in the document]
arminpatel: These were not included in serializer used for ListView so I assumed them to be write only

api/api/clubs/views.py
```diff
@@ -7,3 +8,9 @@
 class ListClubsView(ListAPIView):
     queryset = Club.objects.all()
     serializer_class = ClubSerializer
+
+
+class UpdateClubsView(UpdateAPIView):
+    permission_classes = (IsAdminUser,)
```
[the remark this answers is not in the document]
arminpatel: Not sure if any user should have permission to update them, should they?

api/api/clubs/serializers.py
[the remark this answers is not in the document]
Samy-33: These should be included in `List` response.

Unrelated to that, Update response should always contain the whole updated document.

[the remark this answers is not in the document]
arminpatel: > Unrelated to that, Update response should always contain the whole updated document.

Ok ... Any reason for that?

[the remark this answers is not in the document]
Samy-33: According to: https://google.aip.dev/134

> The response should include the fully-populated resource, and must include any fields that were sent and included in the update mask unless they are input only

The full document in the response helps the clients to parse information from the response itself.

The client would have to query the db again for any fields not present in the `request` that were updated. For example:
  - maybe client wants to use `updated_at` or `updated_by` fields which won't be available in the `response`. Client has to send a GET request to the API and get these fields.

api/api/clubs/views.py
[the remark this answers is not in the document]
Samy-33: Position holders of the club should also be allowed to update the club. Right?
cc @Aksh-Bansal-dev @garg3133

api/api/clubs/tests.py
```diff
@@ -55,3 +58,98 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+        Roles.objects.create(name='Coordinator',
+                             club=test_club,
+                             user=user,
+                             assigned_at=date.today())
+        # when
+        response = self.client.put(f'/clubs/{test_club.id}/', {'name': 'changed_club_name',
+                                                               'category': "S&T",
+                                                               'description': "Some desc here",
+                                                               'email': "theprogclub@iiitdmj.ac.in",
+                                                               'logo': "https://iiitdmj.ac.in"})
```
Samy-33, line 146: Is this update required or the below one is enough?

api/api/clubs/tests.py
```diff
@@ -55,3 +58,98 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+
+        # then
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    @pytest.mark.django_db
+    def testUpdateClubs_positionHolderRequest_updatesSuccessful(self):
```
Samy-33, line 89: Use `pytest.mark.parametrize` to test for all the position holders.

api/api/clubs/tests.py
```diff
@@ -55,3 +58,98 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+                                         'name': 'changed_club_name',
+                                         'category': 'S&T',
+                                         'logo': 'https://iiitdmj.ac.in',
+                                         'description': 'Some desc here',
+                                         'email': 'theprogclub@iiitdmj.ac.in'})
+        self.assertEqual(updated_club.name, 'changed_club_name')
+        self.assertEqual(updated_club.category, 'S&T')
+
+    @pytest.mark.django_db
+    def testUpdateClubs_incompleteInformationInRequest_throwsBadRequest(self):
```
Samy-33, line 125: From the name it's not clear what information is not provided. Maybe a document it at `when`

api/api/clubs/views.py
```diff
@@ -1,9 +1,16 @@
...
 from .serializers import ClubSerializer
 
 
 class ListClubsView(ListAPIView):
     queryset = Club.objects.all()
     serializer_class = ClubSerializer
+
+
+class UpdateClubsView(UpdateAPIView):
+    permission_classes = (IsPositionHolder,)
```
Samy-33, line 14: We should also allow admin user to update clubs.

api/api/clubs/permissions.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class IsPositionHolder(permissions.BasePermission):
+    """Custom permission to allow only position holders of respective club"""
+
+    def has_object_permission(self, request, view, obj):
+        allowed_roles = obj.roles.filter(Q(name='Coordinator') | Q(name='Co-Coordinator'))
+
+        # return true when allowed_roles and users roles have common elements
+        return (allowed_roles & request.user.roles.all()).exists()
```
Samy-33, line 12: A small optimization: instead of considering all the roles of the user, we could consider just the roles of the user for that club.
```python
request.user.roles.filter(club__id=obj.id)
```

Just try it out.

Samy-33 (changes requested):

api/api/clubs/tests.py
```diff
@@ -55,3 +58,133 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+        response = self.client.put(f'/clubs/{test_club.id}/', {'name': 'changed_club_name',
+                                                               'category': "S&T",
+                                                               'description': "Some desc here",
+                                                               'logo': "https://iiitdmj.ac.in"})
+
+        # then
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+
+# function based tests
```
Samy-33, line 117: Unnecessary?

api/api/clubs/tests.py
```diff
@@ -55,3 +58,133 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+
+
+updated_club_data = [('name', 'changed_club_name'),
```
Samy-33, line 120: Should not be present as the global variable. Why?

api/api/clubs/tests.py
```diff
@@ -55,3 +58,133 @@ def testListClubs_hundredClubExistsInDb_returnsFiftyClubInFirstResponse(self):
...
+                     ('category', 'Cultural'),
+                     ('description', 'some_desc_here'),
+                     ('email', 'changed@email.address'),
+                     ('logo', 'https://changedlogo.com')]
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('field, value', [
```
Samy-33, line 128: I think this is not a good use of `parametrize`. For multiple input cases, you should use it.
Here the whole test would be run just to assert a different field value pair again.

You could simply assert each of these fields in the test.

api/api/clubs/permissions.py
```diff
@@ -0,0 +1,13 @@
+from django.db.models import Q
+from rest_framework import permissions
+
+
+class IsPositionHolderOrAdmin(permissions.BasePermission):
+    """Custom permission to allow only position holders of respective club"""
```
Samy-33, line 6: + and admin

Samy-33 (changes requested):

api/api/clubs/tests.py
arminpatel: Is there a better way to use parametrize in this case then? 🤔

arminpatel: Sorry misunderstood your last review. I'll make necessary changes.

Samy-33 (approved):
LGTM

PR #182, opened 2022-07-20
fix(api): add user.id to auth response

SahanY099:
- Changed TokenUserSerializer fields to include user id.


Samy-33:
Squash both commits and fix the message.

Samy-33 (approved):
LGTM

PR #188, opened 2022-07-25
feat(api): configure logging

arminpatel:


api/api/settings/settings.py
```diff
@@ -94,6 +94,26 @@
...
+    },
+    'handlers': {
+        'console': {
+            'class': 'logging.StreamHandler',
+            'formatter': 'default',
+        },
+    },
+    'root': {
+        'handlers': ['console'],
+        'level': 'DEBUG',
```
Samy-33, line 114: Let's keep env specific values in env specific settings

Samy-33 (changes requested):

Samy-33 (approved):
LGTM

PR #195, opened 2022-08-06
feat(api): add endpoint to list and create events

arminpatel:
Event create access is given to `core members`, `co-coordinators`, and `coordinators`.

### DIFF
--- a/api/api/auth/models.py
+++ b/api/api/auth/models.py
@@ -21,7 +21,7 @@ class Token(models.Model):
     @staticmethod
     def if_expired_get_new(token, user):
         token_life = (datetime.now(timezone.utc) - token.created_at).days
-        if(token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS):
+        if (token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS):
             token.delete()
             return Token.objects.create(token=uuid.uuid4().hex, user=user)
         return token
--- a/api/api/auth/models.py
+++ b/api/api/auth/models.py
@@ -21,7 +21,7 @@ class Token(models.Model):
     @staticmethod
     def if_expired_get_new(token, user):
         token_life = (datetime.now(timezone.utc) - token.created_at).days
-        if (token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS):
+        if token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS:
             token.delete()
             return Token.objects.create(token=uuid.uuid4().hex, user=user)
         return token
--- a/api/api/events/test_list_create_events.py
+++ b/api/api/events/test_list_create_events.py
@@ -1,0 +1,173 @@
+import pytest
+from api.clubs.models import Club
+from api.events.models import Events
+from api.roles.models import Roles
+from django.contrib.auth import get_user_model
+from rest_framework import status
+from rest_framework.test import APIClient
+
+User = get_user_model()
+
+
+@pytest.mark.django_db
+@pytest.fixture(scope='function')
+def test_user(django_user_model):
+    return django_user_model.objects.create(email='test@user.com',
+                                            first_name='test',
+                                            last_name='User')
+
+
+@pytest.mark.django_db
+@pytest.fixture(scope='function')
+def test_club(django_user_model):
+    return Club.objects.create(name="Bitbyte - The Programming Club",
+                               category="S&T",
+                               description="Some desc",
+                               email="theprogclub@iiitdmj.ac.in",
+                               logo="https://www.iiitdmj.ac.in/webix.iiitdmj.ac.in/tpclogo.png")
+
+
+@pytest.fixture(scope='function')
+def client():
+    return APIClient()
+
+
+@pytest.mark.django_db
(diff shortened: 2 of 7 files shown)


api/api/auth/models.py
```diff
@@ -21,7 +21,7 @@ def find_or_create(user):
     @staticmethod
     def if_expired_get_new(token, user):
         token_life = (datetime.now(timezone.utc) - token.created_at).days
-        if(token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS):
+        if (token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS):
```
Samy-33, line 24: (...) not required.

api/api/events/test_list_create_events.py
```diff
@@ -0,0 +1,173 @@
...
+@pytest.fixture(scope='function')
+def test_user(django_user_model):
+    return django_user_model.objects.create(email='test@user.com',
+                                            first_name='test',
+                                            last_name='User')
+
+
+@pytest.mark.django_db
+@pytest.fixture(scope='function')
+def test_club(django_user_model):
```
Samy-33, line 22: `django_user_model` is not being used.

api/api/events/test_list_create_events.py
```diff
@@ -0,0 +1,173 @@
...
+    response = client.get('/events/', format='json')
+
+    # then
+    assert response.status_code == status.HTTP_200_OK
+    assert response.data['count'] == 1
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('position', ['Core member', 'Coordinator', 'Co-Coordinator'])
+def testCreateEvents_coreMemberRequest_creationSuccessful(client, test_user, test_club, position):
```
Samy-33, line 70: Should be `testCreateEvent` at all the other places as well.

api/api/events/test_list_create_events.py
```diff
@@ -0,0 +1,173 @@
...
+                             'club': 1,
+                             'registration_approval_required': False,
+                             'address': 'c'}
+    assert created_event.name == 'Introduction to Git and Github'
+
+
+@pytest.mark.django_db
+def testCreateEvents_adminRequest_creationSuccessful(client, test_user, test_club):
+    # given
+    test_user.is_staff = True
```
Samy-33, line 109: After update, don't we need to save the test_user?

api/api/events/test_list_create_events.py
```diff
@@ -0,0 +1,173 @@
...
+                                        'image_url': 'https://asite.com',
+                                        'address': 'c',
+                                        'club': test_club.id
+                                        })
+    # then
+    assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+@pytest.mark.django_db
+def testCreateEvents_missingDataInRequest_throwsBasRequest(client, test_user, test_club):
```
Samy-33, line 159: `throwsBadRequest`

api/api/events/test_list_create_events.py
```diff
@@ -0,0 +1,173 @@
...
+    # request missing start and end time
+    response = client.post('/events/', {'name': 'Introduction to Git and Github',
+                                        'description': 'some desc',
+                                        'image_url': 'https://asite.com',
+                                        'address': 'c',
+                                        'club': test_club.id
+                                        })
+
+    # then
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
Samy-33, line 173: Should we also test, what the error message is? It should clearly state what's missing.

Samy-33 (changes requested):

Samy-33:
Try to break down features into multiple PRs. That way it's easier to review and also one can perform a better review.

api/api/events/test_list_create_events.py
arminpatel: > After update, don't we need to save the test_user?

It does needs to be saved according to the docs.

api/api/events/test_list_create_events.py
```diff
@@ -0,0 +1,182 @@
...
+    # when
+    response = client.get('/events/', format='json')
+
+    # then
+    assert response.status_code == status.HTTP_200_OK
+    assert response.data['count'] == 1
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('position', ['Core member', 'Coordinator', 'Co-Coordinator'])
```
Samy-33, line 69: Lets use constants defined for these values i.e. `Role.CORE_MEMBER` and so on.

api/api/events/views.py
```diff
@@ -0,0 +1,27 @@
...
+        if self._has_create_permission(serializer.validated_data):
+            serializer.save()
+        else:
+            raise PermissionDenied({"message": "You are not allowed to perform this action"})
+
+    # Allow only Core members (including Coordinator and Co-Coordinator)
+    # and admins to create events
+    def _has_create_permission(self, validated_data):
+        allowed_roles = validated_data['club'].roles.filter(
+                        Q(name='Core member') | Q(name='Coordinator') | Q(name='Co-Coordinator'))
```
Samy-33, line 24: Same, let's use constants defined for these roles.

Samy-33 (changes requested):

Samy-33 (approved):
LGTM

PR #196, opened 2022-08-07
feat(api): add endpoint to get user info

arminpatel:


api/api/auth/models.py
```diff
@@ -21,7 +21,7 @@ def find_or_create(user):
     @staticmethod
     def if_expired_get_new(token, user):
         token_life = (datetime.now(timezone.utc) - token.created_at).days
-        if(token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS):
+        if (token_life >= settings.AUTH_ALLOWED_TOKEN_LIFE_IN_DAYS):
```
Samy-33, line 24: surrounding parenthesis not required.

api/api/accounts/permissions.py
```diff
@@ -5,4 +5,6 @@ class IsOwnerOrAdmin(permissions.BasePermission):
     """Custom permission to allow only Owners of profile and Admin Users"""
 
     def has_object_permission(self, request, view, obj):
+        if request.method in permissions.SAFE_METHODS:
```
Samy-33, line 8: This is something different then what the permission name suggests. Why do we need this?

Samy-33 (changes requested):

api/api/accounts/permissions.py
arminpatel: > Why do we need this?

I added this so that any user can retrieve any other user's profile. 

> This is something different then what the permission name suggests.

Hmm. I'll add another permission.

Samy-33: > I added this so that any user can retrieve any other user's profile.

Do we have a usecase for that yet?

arminpatel: I assumed there to be 🤔 . @Aksh-Bansal-dev ?

Aksh-Bansal-dev: I don't think we need this right now but maybe, in the future, we can allow club officials to view members' profiles.

Samy-33 (approved):
LGTM

PR #201, opened 2022-08-12
feat(api): add endpoint to get list of clubs in which the user is a c…

PranshuNayak:


Samy-33:
The URL should be `/users/{user_id}/roles` as a `user` is a parent resource for the `roles` that we are fetching.

And this code should be a part of `account` app.

Samy-33:
For a user you could access their roles by: `user.roles`.

PranshuNayak:
> The URL should be `/users/{user_id}/roles` as a `user` is a parent resource for the `roles` that we are fetching.
> 
> And this code should be a part of `account` app.

Something weird is happening with test_url `f'users/{user.id}/roles'` .  See `captured log call` for reference
error
.

arminpatel:
replacing `f'users/{user.id}/roles/'` with `f'/users/{user.id}/roles/'` should work I guess.

PranshuNayak:
> replacing `f'users/{user.id}/roles/'` with `f'/users/{user.id}/roles/'` should work I guess.

It did thanks!

api/api/accounts/views.py
```diff
@@ -1,22 +1,41 @@
...
+        ROLE_COORDINATOR = 'Coordinator'
+        ROLE_CO_COORDINATOR = 'Co-Coordinator'
+        ROLE_CORE_MEMBER = 'Core member'
+
+        queryset = User.objects.get(pk=pk).roles.all().filter(
+            name__in=[ROLE_COORDINATOR, ROLE_CO_COORDINATOR, ROLE_CORE_MEMBER])
+        if queryset:
+            serializer = RolesSerializer(queryset, many=True)
+            return Response(status=status.HTTP_200_OK, data=serializer.data)
+        return Response(status=status.HTTP_404_NOT_FOUND)
```
Samy-33, line 41: Let's return an empty list instead of 404. Avoid exceptions and unnecessary errors wherever we can.

api/api/accounts/views.py
```diff
@@ -1,22 +1,41 @@
...
     permission_classes = (IsOwnerOrAdmin,)
     queryset = User.objects.all()
     serializer_class = UserSerializer
+
+
+class ListUserRoles(APIView):
+    def get(self, request, pk):
+        ROLE_COORDINATOR = 'Coordinator'
+        ROLE_CO_COORDINATOR = 'Co-Coordinator'
+        ROLE_CORE_MEMBER = 'Core member'
```
Samy-33, line 34: We can use variables defined in `api.roles.model.Role` class.

api/api/accounts/test_list_userroles.py
```diff
@@ -0,0 +1,58 @@
...
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from api.clubs.models import Club
+from api.roles.models import Roles
+
+User = get_user_model()
+
+
+class ListUserRolesTest(APITestCase):
```
Samy-33, line 14: Prefer function based tests. `pytest` is more friendly for function based tests.

Refer: https://github.com/BitByte-TPC/gymkhana/blob/master/api/api/events/test_list_create_events.py

api/api/accounts/test_list_userroles.py
```diff
@@ -0,0 +1,58 @@
...
+    """Tests ListRolesView for getting the list of clubs where user is a core member."""
+    @pytest.mark.django_db
+    def testListRoles_userIsNot_CoreMember_returnsHTTPNotFound(self):
```
Samy-33, line 17: `userIsNotCoreMember` without underscore. Should check for empty list.

api/api/accounts/test_list_userroles.py
```diff
@@ -0,0 +1,58 @@
...
+        club = Club.objects.create(name="Bitbyte - The Programming Club",
+                                        category="S&T",
+                                        description="Some desc here",
+                                        email="theprogclub@iiitdmj.ac.in",
+                                        logo="https://iiitdmj.ac.in")
+        Roles.objects.create(name='Faculty Incharge', user=user,
+                             club=club, assigned_at=date.today(), active=True)
+
+        # when
+
```
Samy-33, line 32: unnecessary blank line.

api/api/accounts/test_list_userroles.py
```diff
@@ -0,0 +1,58 @@
+from datetime import date
+
+import pytest
+from django.contrib.auth import get_user_model
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from api.clubs.models import Club
+from api.roles.models import Roles
```
Samy-33, line 9: We should update the `Roles` class to `Role`, the class represent definition of a single object. Maybe not in this PR.

api/api/accounts/views.py
```diff
@@ -1,22 +1,41 @@
...
+
+        queryset = User.objects.get(pk=pk).roles.all().filter(
```
Samy-33, line 36: nit: Could it just be `roles.filter()` instead of `roles.all().filter()`?

Samy-33 (changes requested):

PranshuNayak:
@Samy-33 made the changes!

PranshuNayak:
> @Samy-33 made the changes!

I think you missed ?

api/api/accounts/test_list_userroles.py
```diff
@@ -0,0 +1,70 @@
...
+    client.force_authenticate(test_user)
+
+    Roles.objects.create(name='Faculty Incharge', user=test_user,
+                         club=test_club, assigned_at=date.today(), active=True)
+
+    # when
+    response = client.get(f'/users/{test_user.id}/roles/', format='json')
+
+    # then
+    assert response.status_code == status.HTTP_204_NO_CONTENT
```
Samy-33, line 49: Should be HTTP_200_OK, with an empty list.

api/api/roles/serializers.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import serializers
+
+from api.clubs.serializers import ClubSerializer
+from api.roles.models import Roles
+
+
+class RolesSerializer(serializers.ModelSerializer):
```
Samy-33, line 7: Should be RoleSerializer.

api/api/roles/serializers.py
```diff
@@ -0,0 +1,12 @@
...
+    club = ClubSerializer(read_only=True)
+
+    class Meta:
+        model = Roles
+        fields = ['name', 'club', 'assigned_at', 'active']
```
Samy-33, line 12: assigned_at should be a read_only field.

api/api/accounts/views.py
```diff
@@ -1,22 +1,38 @@
...
 
-class RetrieveUpdateUserView(generics.RetrieveUpdateAPIView):
+class RetrieveUpdateUserView(RetrieveUpdateAPIView):
     """ View to retrieve and update user """
     permission_classes = (IsOwnerOrAdmin,)
     queryset = User.objects.all()
     serializer_class = UserSerializer
+
+
+class ListUserRoles(APIView):
```
Samy-33, line 31: AFAIU, this won't return paginated response. Would it?

Can using ListAPIView be used to support pagination here? Also, modify tests accordingly.
Using ListAPIView is also a more rest_frameworky way to do this.

Samy-33 (changes requested):

api/api/accounts/views.py
```diff
@@ -1,38 +1,36 @@
...
     queryset = User.objects.all()
     serializer_class = UserSerializer
 
 
-class ListUserRoles(APIView):
-    def get(self, request, pk):
-        queryset = User.objects.get(pk=pk).roles.filter(
+class ListUserRoles(generics.ListAPIView):
+    serializer_class = RoleSerializer
+    lookup_url_kwarg = "pk"
```
Samy-33, line 30: When I read `lookup_url_kwarg`, I don't understand what it means. Can we rename it to something more readable or add a comment over it.

api/api/accounts/views.py
```diff
@@ -1,38 +1,36 @@
...
+
+    def get_queryset(self):
+        pk = self.kwargs.get(self.lookup_url_kwarg)
+        user_roles = User.objects.get(pk=pk).roles.filter(
```
Samy-33, line 34: What happens when the user doesn't exist?

Samy-33 (changes requested):

api/api/accounts/views.py
PranshuNayak: I think comment should work ?

api/api/accounts/views.py
PranshuNayak: We are using
` 'DEFAULT_PERMISSION_CLASSES': (
        'rest_framework.permissions.IsAuthenticated',
    )`
   
So user will be a valid user

api/api/accounts/views.py
```diff
@@ -26,7 +26,9 @@ class RetrieveUpdateUserView(generics.RetrieveUpdateAPIView):
 
 
 class ListUserRoles(generics.ListAPIView):
+    """ View to retrieve user roles where user is a core member of a club """
     serializer_class = RoleSerializer
+    """User id parameter passed as pk to retrieve user roles """
```
Samy-33, line 31: It can be a line comment: """ -> #

PranshuNayak: I will fix it . Moreover I have posted a error on discord . Here is a copy of the same. I have tried deleted migrations and database , recreate them, but it didn't helped
error
.

Samy-33: In my local, for master all the tests are passing. There must have been some change in your PR that introduced some bug. Do send away the full logs

PranshuNayak: Yes! maybe i should make a new pr

PR #204, opened 2022-08-14
feat(api): add endpoint to retrieve, update and delete events

arminpatel:


api/api/events/views.py
```diff
@@ -29,3 +30,10 @@ def _has_create_permission(self, validated_data):
         user_roles = self.request.user.roles.filter(club__id=validated_data['club'].id)
 
         return (self.request.user.is_staff or (allowed_roles & user_roles).exists())
+
+
+class RetrieveUpdateDestroyEventsView(RetrieveUpdateDestroyAPIView):
```
Samy-33, line 35: Should `retrieving` be restricted to core members or admin?

arminpatel:
Is there a way then to split this PR into smaller ones?

Samy-33:
> Is there a way then to split this PR into smaller ones?

I am not sure if there's a pragmatic way to do that. But afaik, you can use git rebase to do that.

Or better, if the changes are per file basis: you could git rebase --soft and then choose which files to include in which commit.

arminpatel:
I guess then we can't split this as retrieve, update and destroy have the same code only tests are different.

Samy-33:
> Or better, if the changes are per file basis: you could git rebase --soft and then choose which files to include in which commit.

My bad I meant, `git reset --soft`

Samy-33 (dismissed):

Samy-33 (approved):

PR #205, opened 2022-08-14
test(api): add tests for retrieve events endpoint

arminpatel:


Samy-33:
Let's include tests in the same PR as the feature.

Samy-33:
When I said break down your work in multiple PRs, I meant break feature wise.
For example:
1. You're working on a feature A
2. Feature A can be divided into a, b, c sub tasks 
3. You raise a PR for task a and its tests
4. You raise a PR for task b and its tests
5. You raise a PR for task c and its tests

PR #213, opened 2022-08-16
chore(ui): add generics in useAuthFetch hook

Aksh-Bansal-dev:


ui/src/components/ClubsCarousel/index.tsx
```diff
@@ -17,15 +18,9 @@ export const CLUB_OPTIONS = [
...
-  name: string;
-  logo: string;
-  category: string;
-}
-
 export const ClubsCarousel: React.FC = () => {
   const [selectedOption, setSelectedOption] = useState(CLUB_OPTIONS[0]);
-  const {data, error} = useAuthFetch('/clubs');
+  const {data, error} = useAuthFetch<{results: ClubData[]}>('/clubs');
   const clubsData = data ? data.results : null;
```
Samy-33, line 24: This is a paginated response, correct?

We can create a `PaginatedResponse<T>` interface, which will contain all the generic paginated related fields as well as an array of Ts.

Aksh-Bansal-dev: What is the type of `next` & `previous` here?
```
count: 3
next: null
previous: null
results: (3) [{…}, {…}, {…}]
```

Samy-33: they are strings

ui/src/api/useAuthFetch.ts
```diff
@@ -14,6 +14,16 @@ const fetcher = async (resource: string) => {
   return res.json();
 };
 
-export const useAuthFetch = (resourse: string) => {
-  return useSWR(resourse, fetcher, {refreshInterval: 10000});
+export interface PaginatedResponse<T> {
+  count: number;
+  next: string | null;
+  previous: string | null;
+  results: T;
```
Samy-33, line 21: results should be an array of T, so maybe `T[]`?

ui/src/components/ClubsCarousel/index.tsx
```diff
@@ -17,15 +18,9 @@ export const CLUB_OPTIONS = [
...
-interface ClubsCarouselData {
-  name: string;
-  logo: string;
-  category: string;
-}
-
 export const ClubsCarousel: React.FC = () => {
   const [selectedOption, setSelectedOption] = useState(CLUB_OPTIONS[0]);
-  const {data, error} = useAuthFetch('/clubs');
+  const {data, error} = useAuthFetch<PaginatedResponse<ClubData[]>>('/clubs');
```
Samy-33, line 23: The reason why it should be `PaginatedResponse<ClubData>` instead of `PaginatedResponse<ClubData[]>` is we know paginated response will return a list for sure and the goal of generics is to hide generic features. The goal and requirement aligns.

Samy-33 (approved):

PR #215, opened 2022-08-17
feat(api): add endpoint for club registration

PranshuNayak:


api/api/clubs/serializers.py
```diff
@@ -10,3 +10,21 @@ class Meta:
         extra_kwargs = {'id': {'read_only': True},
                         'slug': {'read_only': True}
                         }
+
+
+class ClubRegistrationRequestSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ClubRegistrationRequests
```
Samy-33, line 17: Make the ClubRegistrations model singular

api/api/clubs/serializers.py
```diff
@@ -10,3 +10,21 @@ class Meta:
...
+
+
+class ClubRegistrationRequestSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ClubRegistrationRequest
+        fields = ['id', 'user', 'club', 'fee_submitted', 'status', 'remark', 'updated_by']
+        extra_kwargs = {'id': {'read_only': True}}
+
+    def update(self, instance, validated_data):
+        updatable_fields = ['fee_submitted', 'status', 'remark', 'updated_by']
```
Samy-33, line 22: updated_by should not be an updatable field. It should be populated on the backend. The user who sends the update request should be added to updated_by.

PranshuNayak: `updated_by` is a `Foreign key` in `ClubRegistrationRequest` so is it possible ?

Samy-33: `updated_by` should not be accepted as an input from the client. It should be set while processing the request in the view or serializer.

PR #216, opened 2022-08-18
feat(api): add endpoint for roles

arminpatel:


api/api/roles/serializers.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from .models import Roles
+
+
+class RolesSerializer(serializers.ModelSerializer):
```
Samy-33, line 6: Let's keep it singular

Samy-33: Also can you please rename `Roles` model to `Role`

api/api/roles/serializers.py
```diff
@@ -0,0 +1,9 @@
...
+    class Meta:
+        model = Roles
+        fields = ['name', 'club', 'user', 'assigned_at', 'active']
```
Samy-33, line 9: `assigned_at` should be auto populated?

api/api/roles/views.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework.generics import CreateAPIView
+from rest_framework.permissions import IsAdminUser
+
+from .models import Roles
+from .serializers import RolesSerializer
+
+
+class CreateRolesView(CreateAPIView):
```
Samy-33, line 8: CreateRoleView

Samy-33 (changes requested):

api/api/roles/tests.py
```diff
@@ -0,0 +1,80 @@
...
+                                  Role.ROLE_CORE_MEMBER,
+                                  Role.ROLE_FACULTY_INCHARGE,
+                                  Role.ROLE_CONVENER,
+                                  Role.ROLE_CO_CONVENER,
+                                  Role.ROLE_COUNSELLOR,
+                                  Role.ROLE_ASSOCIATE_COUNSELLOR])
+def testCreateRole_adminRequest_returnSuccesful(role):
+    # given
+    client = APIClient()
+    test_user = User.objects.create(email='test@user.com',
```
Samy-33, line 52: I don't understand how these test are succeeding without saving test_user?

arminpatel: > I don't understand how these test are succeeding without saving test_user?

No idea. My bet guess is due to `force_authenticate`.

Samy-33: Hmm, I'll check it out.,

Samy-33 (approved):

Issue #198, opened 2022-08-10
[API] Endpoint to get list of clubs in which the user is a core/co-coordinator/coordinator


Aksh-Bansal-dev:

PranshuNayak:
May be we will need all roles for a user?

Aksh-Bansal-dev:
> May be we will need all roles for a user related to a club?

We need this for #199. If we request roles for an individual club, then it will take 20-30 requests to display this section. So, I think it will be better to request all in a single request.

PranshuNayak:
![error](https://user-images.githubusercontent.com/86283785/184295702-2f44b488-02a7-4b9a-b2ad-77333ebf3e40.jpg)
@Samy-33

PR #221, opened 2022-08-23
feat(api): add endpoint to get list of roles where user is a core member

PranshuNayak:


api/api/accounts/urls.py
```diff
@@ -5,4 +5,5 @@
 urlpatterns = [
   path('', views.CreateUserView.as_view()),
   path('<int:pk>/', views.RetrieveUpdateUserView.as_view()),
+  path('roles/', views.ListUserRoles.as_view()),
```
Samy-33, line 8: url should be `<int:pk>/roles/`

And should return roles for the user with pk in the url path.

Samy-33 (changes requested):
Overall looks good. One small change.

api/api/accounts/urls.py
PranshuNayak: Is this even needed ? Current implementation get user if from `request.user` which is more reliable. We can do this but it would cause more exceptions to handle.

api/api/accounts/views.py
```diff
@@ -20,3 +23,14 @@ class RetrieveUpdateUserView(generics.RetrieveUpdateAPIView):
     permission_classes = (IsOwnerOrAdmin,)
     queryset = User.objects.all()
     serializer_class = UserSerializer
+
+
+class ListUserRoles(generics.ListAPIView):
```
Samy-33, line 28: Correct me if I am wrong. This endpoint returns roles of the users where they are core members of a club.
What's stopping us from making it a generic endpoint that returns all the user roles?

Moreover if we are not using user-id from url path, it doesn't make sense to have url like: `/accounts/<int:pk>/roles/`, let's just have it like: `/accounts/me/roles/`

api/api/roles/serializers.py
```diff
@@ -8,3 +8,10 @@ class Meta:
         model = Role
         fields = ['name', 'club', 'user', 'assigned_at', 'active']
         extra_kwargs = {'assigned_at': {'read_only': True}}
+
+
+class ListUserRolesSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Role
+        fields = ['name', 'club', 'assigned_at', 'active']
```
Samy-33, line 16: Currently `club` would be an integer representing `id` of a club. Do we need any more information about the club in the response?

PranshuNayak: @Aksh-Bansal-dev name would be fine?

api/api/accounts/views.py
PranshuNayak: @Aksh-Bansal-dev this works?

api/api/roles/serializers.py
Aksh-Bansal-dev: Yeah, we only need the clubs' names.

api/api/accounts/views.py
Aksh-Bansal-dev: I think `/accounts/me/roles/` would be fine. We can just get the user-id from the token.

PranshuNayak: Not about the url ... Returning all roles is okay or is this specific to roles like (co,co-co,and core) as mentioned in the issue #198

Aksh-Bansal-dev: Yeah, returning all roles would be better.

PranshuNayak:
@Samy-33 made the changes

api/api/accounts/test_list_userroles.py
```diff
@@ -60,8 +58,13 @@ def testListRoles_userIsCoreMember_returnsFirstFiftyRolesList(client, test_user,
...
-    response = client.get(f'/users/{test_user.id}/roles/', format='json')
+    response = client.get('/users/accounts/me/roles/', format='json')
 
     # then
     assert response.status_code == status.HTTP_200_OK
     assert len(response.data['results']) == 50
+    assert response.data['results'][0] == {'name': 'Coordinator',
+                                           'club__name': 'Bitbyte - The Programming Club',
+                                           'club__slug': 'slug field',
+                                           'assigned_at': '2022-08-31',
```
Samy-33, line 69: wouldn't it fail on 1st September 2022?

api/api/roles/serializers.py
```diff
@@ -11,7 +11,10 @@ class Meta:
 
 
 class ListUserRolesSerializer(serializers.ModelSerializer):
+    club__name = serializers.CharField(max_length=100)
```
Samy-33, line 14: is there a reason why double underscores are used?

a better response would be
```
role: {
...
  club: {
    name: ...,
    slug: ...
  }
...
}```

Use ClubSerializer to list the club related fields.

PranshuNayak: Instead of serializing 'foreign key club' I was extracting 'club's name & slug' that was needed so using double underscores. Current `ClubSerializer` would return other details too so I thought this was better. Will modify this anyway

api/api/accounts/test_list_userroles.py
PranshuNayak: Why ?

api/api/roles/serializers.py
Samy-33: Will current approach return the format I mentioned?

api/api/accounts/test_list_userroles.py
Samy-33: While creating the role, you're setting `assigned_at`  by using date.today(), which will be different on different days. But you're still asserting it with 31st August 2022.

api/api/roles/serializers.py
PranshuNayak: No it won't . But `ClubSerializer` will return fields other than `name & slug` too , so we need to define other custom serializer

Samy-33: No doesn't strictly need to return name and slug. You can return other club fields along as well.

PranshuNayak:
Ready for review

PranshuNayak:
Pending for review

PR #230, opened 2022-08-31
feat(ui): added feature where events are fetched from api

a1pharooster:


a1pharooster:
@Aksh-Bansal-dev  can you review this

